PR #14, opened 2019-04-27
fix typo

ousttrue:
I have tried this plugin. With this fix, `install_gadget.py` worked.

* Windows10
* NVIM v0.4.0-533-g052ced495


puremourning:
Thanks!

Just so you know, I'm fairly confident that the plugin doesn't generally work on Windows, but I'm willing to accept patches to make it so, though I'm not likely to test it on windows in the faceable future.

It also doesn't work with NeoVim. Support for NeoVim is low on the priority list however.

PR #15, opened 2019-04-28
Add ClearBreakpoints API

puremourning:
This is mainly to set up testing properly and test the actual mappings rather than just that mappings exist.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/15)

PR #17, opened 2019-04-28
run_tests should build the test programs

puremourning:
rather than only in CI. This means you just ./run_tests to run them locally.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/17)

PR #21, opened 2019-05-08
Test clear breakpoints 

puremourning:
And generally fix the tests so that they test something.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/21)

PR #22, opened 2019-05-16
Removing missing images and spell correction

adelarsq:
Removing missing images and a little spell correction.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/22)

PR #31, opened 2019-06-01
Update servers

puremourning:
Update the server versions


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/31)

PR #32, opened 2019-06-01
Make the tests actually do some debugging

puremourning:
Actually fire up the debugger and make a step. Prove that the PC gets updated correctly and the current line/col are also updated.

This involved pulling in more of vim's test code, so that was all moved into a `tests/lib` directory


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/32)

Issue #35, opened 2019-06-19
[Feature Request]: Support C# Samsung debugger


jpfeiffer16:
I'd be interested in adding C# support with the new Samsung debugger: https://github.com/Samsung/netcoredbg

I know there were discussions about it using the Microsoft vsdbg, but that didn't go anywhere because of licensing issues.

@puremourning do you know anything about this? I feel like I saw a pr or something that added that support and I was trying to find that to go off of, but use the Samsung debugger instead.

puremourning:
In theory you should just be able to configure it up in vimspector.json   

The docs on this aren’t exactly stellar but essentialy you just have to get it jnstallled somewhere and find out how to start it, then work out the launch parameters from its package.json.

puremourning:
From looking at the code (sigh) the launch arguments seem to be:  

- program (string)
- args (list)
- stopOnEntry (bool)

https://github.com/Samsung/netcoredbg/blob/master/src/debug/netcoredbg/vscodeprotocol.cpp#L316

And the README says how to build and run it. 

So should be easy without vimspector changes.

puremourning:
> ? I feel like I saw a pr or something that added that support

Possibly because install_gadget includes (disabled) ability to install the one with issues. I should squelch that code.

jpfeiffer16:
Ok. I'll give that a try! Sorry for the ignorance, what is install_gadget?

jpfeiffer16:
Oh. I see the python script.

puremourning:
oh install_gadget.py is the installer script that downloads and installs some debug adapters: https://github.com/puremourning/vimspector/blob/master/install_gadget.py

If you can get it working, i'd be thrilled to merge a pr which includes it and updates the readme to say it's supported.

puremourning:
I pushed a branch (c-sharp-samsung) which adds the adapter to install_gadget.py

I haven't actually tested that it works because i don't really have a trivial c-sharp (botnet core) application i could test quickly on macOS.

You _should_ be able to use it by running './install_gadget.py --force-enable-csharp' with a .vimspector.json like this:

```json
{
    "configurations": {
        "launch": {
            "adapter": "netcoredbg",
             "configuration": {
                 "request": "launch", 
                 "program": "${workspaceRoot}/path/to/binary",
                 "args": [ "arg1", "arg2...." ],
                 "stopAtEntry": true 
             }
        }
    }
}
```

let me know how you get on.

jpfeiffer16:
[...]

PR #36, opened 2019-06-22
C sharp samsung

jpfeiffer16:
@puremourning, these are my changes for #35.

I made three changes here:
1. When doing requests to lookup variable requests, if the individual variable response elements do not have an 'expensive' key just assume they are not. Currently it is throwing a key error there.
2. Fixed what I think was probably a typo in the event handler for `OnEvent_capabilities`that was making so the capabilities never got set and were not respected afterwards. In this case it was `supportsConfigurationDoneRequest`.
3. I added back in the mono debugger adapter. I know I mentioned this earlier and that might have led to some confusion. I don't think you need to remove this as this is a openly licensed debug adapter (I believe. I haven't done extensive research though so correct me if I'm wrong.) I checked this back in under the impression that I misspoke and that  you removed it thinking it was the proprietary  microsoft dotnet runtime debugger. If this is not the case and you had another reason, I will gladly revert the commit. Sorry for the confusion here.

### DIFF
--- a/install_gadget.py
+++ b/install_gadget.py
@@ -142,23 +142,6 @@ GADGETS = {
       },
     }
   },
-  'vscode-mono-debug': {
-    'language': 'csharp',
-    'enabled': False,
-    'download': {
-      'url': 'https://marketplace.visualstudio.com/_apis/public/gallery/'
-             'publishers/ms-vscode/vsextensions/mono-debug/${version}/'
-             'vspackage',
-      'target': 'vscode-mono-debug.tar.gz',
-      'format': 'tar',
-    },
-    'all': {
-      'file_name': 'vscode-mono-debug.vsix',
-      'version': '0.15.8',
-      'checksum':
-          '723eb2b621b99d65a24f215cb64b45f5fe694105613a900a03c859a62a810470',
-    }
-  },
   'vscode-bash-debug': {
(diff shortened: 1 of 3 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/36)

jpfeiffer16:
I'm trying to run the tests locally to check the failures but I'm getting an error on the `which` command in the tests script `./run_tests`. Illegal option -s

nosami:
The Mono debugger adapter _is_ OSS

puremourning:
Thanks for the PR!

FWIW, There’s a docker image that you can use to run the tests locally in the same image that azure runs (on Linux). 

Looks like there were CI failures.

puremourning:
Regarding the c# debug adapter, when I looked before I found there were licensing restrictions for one and I couldn’t make another work. I’m not precious about any particular adapter. I don’t actually use c# so provided there’s no license conflict im happy to go with any that works best. We use omnisharp in ycmd so that seems fine.

puremourning:
Thanks for the PR! I used your commits and added a couple of my own (docs, minimal test so I can check it)

jpfeiffer16:
No problem! Thanks for using it.

PR #37, opened 2019-06-23
C sharp samsung

puremourning:
Closes #36 

With @jpfeiffer16's help, adding c-sharp support, including:

- updated the docs
- added a minimal test app


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/37)

PR #42, opened 2019-07-05
Fix and test breakpoint issues

### LINKED ISSUE #12: Can't disable breakpoint while debugging
Since moving breakpoints code around, you can no longer disable a breakpoint while actually debugging.

* Set a breakpoint (F9)
* Start debugging (F5)
* When t he breakpoint triggers, F9 again on the line
* continue (F5)

expected: Doesn't trigger again
actual: triggers again.

NOTE: Can still disable, delete when _not_ debugging.
NOTE: secondary issue is that PC and BP symbols overlap, though I don't know of a solution for that

puremourning:
Make the test framework a vim plugin (sort of)
Refactor some common parts.
Finally add some solid tests for breakpoint handling both at runtime and not.
Fix the issues that came out of that.

Purpose is to be able to properly refactor this stuff when we have good test coverage.

Fixes #12


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/42)

puremourning:
Seems ok, but the tests are super flaky.

- increase timeouts ?
- investigate the "refer to deleted buffer" issue on resetting.

PR #45, opened 2019-07-26
Allow dependent groups of variables in a list

puremourning:
Allows for variables to reference other variables in the same layer:

```json
+      "variables": [
+        {
+          "python": {
+            "shell": "/bin/bash -c 'if [ -z \"${VIRTUAL_ENV}\" ]; then echo $$(which python); else echo \"${VIRTUAL_ENV}/bin/python\"; fi'"
+          },
+          "nosetests": {
+            "shell": "/bin/bash -c 'if [ -z \"${VIRTUAL_ENV}\" ]; then echo $$(which nosetests); else echo \"${VIRTUAL_ENV}/bin/nosetests\"; fi'"
+          }
+        },
+        {
+          "python_path": {
+            "shell": [
+              "${python}",
+              "${workspaceRoot}/run_tests.py",
+              "--dump-path"
+            ]
+          }
         }
```

### DIFF
--- a/python3/vimspector/utils.py
+++ b/python3/vimspector/utils.py
@@ -33,6 +33,10 @@ def SetUpLogging( logger ):
     logger.addHandler( _log_handler )
 
 
+_logger = logging.getLogger( __name__ )
+SetUpLogging( _logger )
+
+
 def BufferNumberForFile( file_name ):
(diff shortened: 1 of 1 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/45)

PR #49, opened 2019-07-31
Go support using vscode-go

### LINKED ISSUE #48: Go support via Delve
_If you will excuse me not completing the template, this is a rather more vague request for information_

In a tools session at  GopherCon 2019 ([notes](https://docs.google.com/document/d/1-RVyttQ0ncjCpR_sRwizf-Ubedkr0Emwmk2LhnsUOmE/edit#heading=h.4a2ftcuzucfc)) there was an interesting discussion started about Delve and the Debugger Adapter Protocol. Delve has become the de facto debugger for Go, and it looks like it will be blessed as the official debugger by the Go team. 

As part of trying to understand how Delve can become better integrated in various editors, the Debugger Adapter Protocol was brought up. Delve does not currently speak this protocol, but if there is sufficient support the work could be done (relatively easily). 

I've previously seen this project, but it was this discussion that prompted me to raise an issue:

* what are your thoughts/experience of the Debugger Adapter Protocol?
* do you have any advice on how to proceed on the question of how to integrate Delve with Vim (and other editors)
* ...

The context of considering the Debugger Adapter Protocol is the Go now has an official LSP server, `gopls`. Hence, building on that success, it seems logical to consider the Debugger Adapter Protocol.

Thanks in advance

puremourning:
Experimental for now.

Fixes #48

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ on a best-efforts basis:
 
 - Java (see caveats)
 - C# (c-sharp) using dotnet core
+- Go (requires separate installation of [Delve][])
 
 ## Languages known not to work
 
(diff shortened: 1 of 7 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/49)

PR #56, opened 2019-10-01
RunInTermal: Re-use a completed terminal

### LINKED ISSUE #54: [Feature Request]: Don't open a new terminal window after every restart
I'm sorry for the issues bombard, I was just thinking that these are different topics to discuss so I didn't want to put them all in one big issue.

So today, after some debug sessions with vimspector, I noticed that the stdout/stderr window of my running program is being recreated every time I start a debug session. I must say that it's kind of annoying to quit the old window every time a new session is started. I'm also a Neovim refugee because of #30 and so I'm not used yet to Vim's `:terminal`s which makes this process a little bit harder for me TBH..

puremourning:
Fixes #54 

As discussed on #54, we can re-use the terminal window if the terminal within it has completed. This makes the common use-case of restarting debugging involve fewer irritating dead terminals in increasingly narrow vsplits.

cc: @doronbehar

### DIFF
--- a/python3/vimspector/code.py
+++ b/python3/vimspector/code.py
@@ -25,6 +25,9 @@ class CodeView( object ):
   def __init__( self, window ):
     self._window = window
 
+    self._terminal_window = None
+    self._terminal_buffer_number = None
+
     self._logger = logging.getLogger( __name__ )
     utils.SetUpLogging( self._logger )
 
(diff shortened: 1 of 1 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/56)

Issue #53, opened 2019-09-30
[Feature Request]: Why are all of those questions when starting a debug session?


doronbehar:
So today, when I've started my first ever debug session I was asked:

```
Enable exception filter 'Raised Exceptions'? (Y/N)
Enable exception filter 'Uncaught Exceptions'? (Y/N)
When to break on exception?
1: never
2: always
3: unhandled
4: userHandled
Type number and <Enter> or click with mouse (empty cancels):
```

And only then the debugging has started. Is it possible to make give these answers default answers for the whole project / adapter / all adapters? Say I don't understand what are these questions all about, where could I find a documentation for it?

puremourning:
It’s a bit of a hack to be honest. 

Basically the way DAP is designed makes some assumptions about what UI you have available. I don’t have a UI for the ‘exception breakpoint’ configuration Yet but it’s useful enough that I threw that in so it could be used. Certainly improving this is on the backlog. 

Basically the debug server is asking if you want to break on exceptions. I didn’t make a UI for that yet so it just asks you every time. Usually the defaults are fine and you can just press enter. Rememberinf choices is also on the backlog. 

Each adapter can have different options so I can’t doc them.

doronbehar:
> I don’t have a UI for the ‘exception breakpoint’ configuration [...]

Ideally, I wouldn't expect a UI for configuring these exception options. I'd expect to be able to set them in `.vimspector.json` for every launcher configuration and to be able to set defaults per adapter in (e.g) `~/.config/nvim/vimspector.json`. Only if none of these will be set, I'd like to be asked and even then not every time.

Maybe the last part of this UX suggestion will be a little bit harder to implement, but I think being able to set defaults in a global / project / launcher configuration would be nice.

puremourning:
> I'd expect to be able to set them in .vimspector.json for every launcher configuration

This might be possible, though the available options are defined as [capabilities returned on the initialize request](https://microsoft.github.io/debug-adapter-protocol/specification#Types_ExceptionBreakpointsFilter) so I don't know for certain how users could know what to configure in advance.

Feel free to have a read of the protocol and propose something :)

> Only if none of these will be set, I'd like to be asked and even then not every time.

yeah like I said, remembering the user's response is on my mental backlog. it's not that hard to do, just requires some code and tests. well, like everything else I suppose.

puremourning:
I was wondering what we did in the tests.

:')

```python
  pyx << EOF
from unittest.mock import patch
[...]

PR #58, opened 2019-10-05
Remember user choices

puremourning:
There are 2 things we ask for input for:

- input variables
- exception breakpoint filters

It's irritating to have to repeat yourself when going through the
edit/debug loop.

Howver, cacheing has some quirks and disadvantages - they key one being
when to clear the cache. To resolve this we take two slightly different
approaches:

1. For input variables, we remember the choice of the user, but present
that only as the default, so they can just hit enter to accept it. We
already rememeber the choices for the length of the debug session (i.e.
across 'restart' calls).

2. For exception breakpoints, we remember the choices for as long as the
current session is running.

This allows users to hit the 'restart' button without being prompted at
all.

Meanwhile, we also remove the (broken) support for exception breakpoint
matchers and state the server default for exception breakpoint filters.

---

Fixes https://github.com/puremourning/vimspector/issues/53 (mostly). There's no way to set exception breakpoint info in vimspector.json currently. We could add that in a separate PR maybe.

cc @doronbehar.

### DIFF
--- a/python3/vimspector/breakpoints.py
+++ b/python3/vimspector/breakpoints.py
@@ -43,7 +43,7 @@ class ProjectBreakpoints( object ):
     # These are the user-entered breakpoints.
     self._line_breakpoints = defaultdict( list )
     self._func_breakpoints = []
-    self._exceptionBreakpoints = None
+    self._exception_breakpoints = None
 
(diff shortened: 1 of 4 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/58)

PR #59, opened 2019-10-06
Support the Node.js debugger (not the legacy one)

puremourning:
Screenshot 2019-10-06 at 19 18 29

There are 2 implementations of this by upstream, but I'm only including the later one.

This allows to debug node.js on the server side. Neat. See also #60 for debugging javascript in the client.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/59)

PR #60, opened 2019-10-06
Add chrome debugger

puremourning:
vimspector-chrome-debug

This allows using vim to debug client code running in chrome, assuming that's a thing you might want to do.

It also irons out a couple of bugs and works around a protocol violation in that server.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/60)

Issue #61, opened 2019-10-12
Python: Attach to process not working with vscode-cpptools


puremourning:
**Describe the bug**

Python attach fails with "Session is not initialised". This appears to be a bug in vscode-cpptools, but requires more investigation. In practice, however, simply attaching directly to the `ptvsd` port works perfectly, so perhaps that should be the recommended approach, i.e.

* Start the application with `python -m ptvsd --host localhost --port 1234 --wait ...`
* Attach to it directly with 

```json
    "python - attach ptvsd": {
      "adapter": {
        "name": "ptvsd",
        "port": 1234
      },
      "configuration": {
        "name": "attach",
        "type": "ptvsd",
        "request": "attach",
        "debugOptions":  [
          "RedirectOutput",
          "UnixClient",
          "ShowReturnValue"
        ]
      }
    }
```

**To Reproduce**

* cd to `support/test/python/simple_python`
* `python -m ptvsd --wait --host localhost --port 1234 test.py`
* In another window, in same directory, set a breakpoint, and fire up vimspector, using the following config:

```json
    "python - attach": {
      "adapter": "vscode-python",
      "variables": [
        {
          "python": {
            "shell": "/bin/bash -c 'if [ -z \"${dollar}VIRTUAL_ENV\" ]; then echo $$(which python); else echo \"${dollar}VIRTUAL_ENV/bin/python\"; fi'"
          }
        }
      ],
      "configuration": {
        "name": "Python attach",
        "type": "vscode-python",
        "request": "attach",

        "pythonPath": "${python}",
        "host": "localhost",
        "port": "${port}"
      }
    },
```

**Expected behavior**

Vimspector starts and attaches to the app

**Actual behaviour**

It never attaches, and the vscode-python node debug adapter spits out an error (which you can see in the console of the `test.py` that the session is not initialized. This is clearly a vscode-python bug and/or incompatibility with ptvsd (it doesn't appear to send a DAP initialise request).

puremourning:
Can no longer reproduce this.

It looks like ptvsd 4 works, but ptvsd 5 does not, which is apparently pre-release according to pypi.

`easy_install` installed `ptvsd 5` for me, but pip knows better. Better use pip next time.

PR #62, opened 2019-10-13
Python ptvsd

puremourning:
There seems to be a bug attaching to processes using vscode-python. But we can connect directly to ptvsd using a socket. This improves socket connections somewhat, by waiting longer (allowing the user more time to start up ptvsd manually)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/62)

puremourning:
ref #61

PR #67, opened 2019-10-27
Don't crash when no file is open on start

puremourning:


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/67)

PR #68, opened 2019-10-27
Source request

puremourning:
`lldb-vscode` can be used directly from vimspector, but it requires the use of `sourceReference` as its break on start is really on start deep in the library.

It turns out that the use of the `source` request and `sourceReference` is actually not optional in the spec. Our solution is simple enough: request when jumping to a frame. 

Yet more cringy race conditions and hacky async requests in this PR, but the result is quiet satisfying:

image

Haven't tried the other adapters yet.

### DIFF
--- a/python3/vimspector/stack_trace.py
+++ b/python3/vimspector/stack_trace.py
@@ -15,12 +15,16 @@
 
 import vim
 import os
+import logging
 
 from vimspector import utils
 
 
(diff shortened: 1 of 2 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/68)

PR #72, opened 2019-12-14
Fix issues with breakpoints not being in the correct location

### LINKED ISSUE #63: Breakpoints: Changing file contents moves signs but does not update breakpoint location
**Describe the bug**

When a breakpoint is set with no active debug session, the sign appears next to the line (OK). If you then insert lines or delete them, the Vim sign moves around with the line (great!), however, Vimspector doesn't know this has happened, and still thinks the breakpoint is on the (absolute) line number which the user specified, leading to strange results and unexpected breakpoints.

**To Reproduce**

* Open some file
* Go to a line
* Press `<F9>`
* Go up a line and delete it
* Try and disbable the breakpoint, (press `<F9>`)

etc., including start debugging.

**Expected behavior**

Breakpoint is disabled, and/or striggers at the expected line.

**Actual behaviour**

Spurious breakpoint created, break on wrong line, etc.

puremourning:
Fixes #63 

If signs are moved by user actions, use the current lnum of the sign for the breakpoint.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -24,7 +24,6 @@ things like:
 
 Along with optional additional configuration for things like:
 
-- Function breakpoints
 - Exception breakpoints
 
 ### Debug adapter configuration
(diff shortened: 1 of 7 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/72)

PR #73, opened 2019-12-15
Syntax highlighting for locals and watches

puremourning:
This enabled syntax highlighting for locals window and watches window. This just aids readability a little. It always uses the syntax of the "code" window, whatever that happens to be.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/73)

PR #75, opened 2020-01-07
Document closing the debugger

orlangure:
Thank you for this project!

After trying this out, I couldn't figure out how to return close the debugger, which turned out to be pretty simple. I suggest to document it as a part of a regular debugging flow.

This PR also includes a couple of other fixes to readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ For a tutorial and usage overview, take a look at the
       * [Variables and scopes](#variables-and-scopes)
       * [Watches](#watches)
       * [Stack Traces](#stack-traces)
-      * [Program Output:](#program-output)
+      * [Program Output](#program-output)
          * [Console](#console)
    * [Debug adapter configuration](#debug-adapter-configuration)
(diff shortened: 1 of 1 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/75)

orlangure:
> Better to link to the mappings section ?

There is no `Reset` in mappings, so I can just omit this suggested mapping (it conflicts with another mapping in human mode). WDYT about keeping just "* `call vimspector#Reset()`"

puremourning:
> There is no Reset in mappings

Ugh, that's weak. I guess i always click Rest button! I'm not even sure it's smart enough to trigger closedown properly if you `:tabclose`.

let's just remove the mapping part for now, and i'll have to add something to the backlog to make this better

puremourning:
Come on bot....

puremourning:
Thanks for the PR!

PR #76, opened 2020-01-07
Make statement about neovim support

puremourning:
Currently unsupported for technical reasons, not philosophical ones.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/76)

PR #80, opened 2020-01-09
Fix: Remove balloon '...' in windows != code_window

tinmarino:
Remove the ... balloon in non code windows

Problem: In non-code window, user see a '...' balloon even if python knows it does not need to work
Solution: use `pyeval` in vim so that python's knowledge is getting back from the stack as return value


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/80)

PR #81, opened 2020-01-09
Feature: Console do not leave insert mode

tinmarino:
Just add a `vim.command('call feedkeys("A", "n")')` after the console appending callback

(and some comments)

### DIFF
--- a/autoload/vimspector.vim
+++ b/autoload/vimspector.vim
@@ -97,6 +97,8 @@ function! vimspector#AddWatchPrompt( expr ) abort
 endfunction
 
 function! vimspector#EvaluateConsole( expr ) abort
+  " Change to normal mode to avoid prompt in output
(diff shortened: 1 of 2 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/81)

puremourning:
Apologies, I haven't got round to looking at this - my bad. I will try it out, i promise.

ayyess:
I also needed to add the feedkeys call to the error handling callback https://github.com/ayyess/vimspector/commit/682b370a11a92820642c14330bd57048f98da7f3#diff-38c8320aa7fa2f4b1a527d1f8074ffd3R150

puremourning:
I think we'll stick with the default behaviour of the prompt buffer. There are other issues taking about improvements to that, but we're kind of limited by what vim provides and what DAP provides.

PR #84, opened 2020-01-10
A selection of misc fixes and enhancements from ongoing usage

puremourning:
* Allow arbitrary ssh arguments, such as specifying a key, or other options
* Fix `frame` being optional 
* Fix go which sometimes fails to work when setting breakpoints in advance
* Fix some issues when servers don't provide `source` or `line` (looking at you java)
* Expand vars in vars. This allows more composition of config.
* Fix true/false when in adapter config's configuration defaults (i.e. don't use json.dumps, use a global)

### DIFF
--- a/.ycm_extra_conf.py
+++ b/.ycm_extra_conf.py
@@ -26,6 +26,15 @@ def Settings( **kwargs ):
             }
           ]
         }
+      },
+      'capabilities': {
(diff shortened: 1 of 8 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/84)

Issue #83, opened 2020-01-10
[Feature Request]: User interface, silent errors if not connected


tinmarino:
**Is your feature request related to a problem? Please describe.**
Press `f6` (Pause) or `f10` (next) without connection gets an error message (here for f10)
```
Error detected while processing function vimspector#StepOver:
line    1:
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home2/tourneboeuf/.vim/pack/bundle/opt/spector/python3/vimspector/debug_session.py", line 306, in StepOver
    if self._stackTraceView.GetCurrentThreadId() is None:
AttributeError: 'NoneType' object has no attribute 'GetCurrentThreadId'
```

**Describe the solution you'd like**
1. Hide the error message with:
```python
if self._stackTraceView is None or self._stackTraceView.GetCurrentThreadId() is None:
```
instead of
```python
self._stackTraceView.GetCurrentThreadId() is None:
```
2. Echo a nice little message `MessageUser` with a warning color asking him to connect first
3. Autoconnect ?

**Describe alternatives you've considered**
1. Do the first stepOver() is auto connecting if not connected ? 
2. Is is preferable to silence the error, and even no message

**Additional context**
My solution would be 1, 2, 3. I'd be glad to make those changes with your desired solution.

puremourning:
Yeah this is a bug. I’ve know about it for ages but been too lazy to fix it. A simple fix is to just make the methods tolerate being called before connection. Im happy with silence or a little user message.

puremourning:
This was fixed in #85 and #77

PR #85, opened 2020-01-10
Check connection before some operation (`next`) to avoid Python errors in vim messages

tinmarino:
Fixing  #83:  Silencing the message with early return from function and little echo warning

__Note:__ I encapsulated the check in a function decorator: This may seem overkill but at first I did not and ended with :

```python
  def StepOver( self ):
    if (not self._connection
        and utils.MessageNoConnection( )
        or self._stackTraceView.GetCurrentThreadId() is None):
      return
    ...
```

```python
  @utils.IfConnected
  def StepOver( self ):
    if self._stackTraceView.GetCurrentThreadId() is None:
      return
    ...
```
As many functions need the check, for readability and avoiding boolean magic, I preferred the second.

PS: I didn't already say, so my groupie's turn:
@puremourning your plugin is awesome !
1. Finally a debugger on vim (of course thk microsoft too)
2. Very well written
    1. __code:__ small functions, sparse: just missing comments other than T.O.D.O and F.I.X.M.E ;-)
    2. __architecture:__ one vim enter, one py enter, one util, everyone in his class. Actually I am still amazed writing these lines because it seems easy once done but thinking of the path, well you got out of troubles spectacularly.
    3. __workflow:__ adding issues, pulls to your own project, and a super test, issue framework on github
3. It works well ! So you already did it ! Congrats.

### DIFF
--- a/python3/vimspector/debug_session.py
+++ b/python3/vimspector/debug_session.py
@@ -20,6 +20,7 @@ import os
 import shlex
 import subprocess
 import traceback
+import functools
 import vim
 
 from vimspector import ( breakpoints,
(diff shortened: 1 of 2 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/85)

tinmarino:
I did not manage the `static` decorator -> I putted it at top of file which is bad. I'll look for a better solution after the WE.

tinmarino:
For the test, sorry, I don't feel very motivated ...

PR #91, opened 2020-01-24
Switch to macOS 10.14 in CI (10.13 will be disabled soon)

puremourning:
Let's see what azure thinks. We know cpptools is broken in 10.15.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/91)

PR #103, opened 2020-01-29
Add a --no-ssl flag that switches off certificate verification

axrt:
Hi @puremourning! 

First of all, thank you so much for the plugin! Man, is it awesome! I thing I had never realized how much struggle it had been to use all kinds of different debuggers (we code in like 5 languages simultaneously)!

This pr is a tiny contribution that might help unfortunate people like myself to bypass the insanity of the certificate chanis that our IT had installed, and which neither JVM (fixed that with some sweat and tears) nor python (no luck after a day of bashing my head against the display) can verify.
Basically, it adds a "--no-ssl" flag which switches off the certificate validation. This helps bypass the self-signed certificates in the chains (in any case, we are downloading from MS's git repo).

I will be more than happy if you choose to accept this pr. Thanks in advance,

axrt

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ README.md.toc.*
 .DS_Store
 *.vimspector.log
 support/test/csharp/*.exe*
+.neomake.log
--- a/install_gadget.py
+++ b/install_gadget.py
@@ -511,7 +511,10 @@ def DownloadFileTo( url, destination, file_name = None, checksum = None, sslchec
   print( "Downloading {} to {}/{}".format( url, destination, file_name ) )
 
   if not sslcheck:
-    kwargs = {"context":  ssl._create_unverified_context()}
+    context = ssl.create_default_context()
+    context.check_hostname = False
+    context.verify_mode = ssl.CERT_NONE
+    kwargs = { "context":  context }
   else:
     kwargs = {}
 
@@ -669,10 +672,9 @@ for name, gadget in GADGETS.items():
            '(when supplying --all)'.format( name, lang ) )
 
 parser.add_argument(
-    "--no-ssl",
+    "--no-check-certificate",
     action = "store_true",
-    help = "Switch the gadget download ssl certificate verification"
-    "off (use only in you have a self-signed certificate in your chain)"
+    help = "Do not verify SSL certificates for file downloads."
 )
 
 args = parser.parse_args()
@@ -706,7 +708,7 @@ for name, gadget in GADGETS.items():
       destination = os.path.join( gadget_dir, 'download', name, v[ 'version' ] )
 
       url = string.Template( gadget[ 'download' ][ 'url' ] ).substitute( v )
-      verify_cert_off = getattr(args, "no_ssl")
+      verify_cert_off = args.no_check_certificate
 
       file_path = DownloadFileTo(
         url,
--- a/install_gadget.py
+++ b/install_gadget.py
@@ -35,6 +35,7 @@ import sys
 import json
 import functools
 import time
+import ssl
 
 try:
   from io import BytesIO # for Python 3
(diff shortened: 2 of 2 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/103)

axrt:
> _[.gitignore, line 18 at r1](https://reviewable.io/reviews/puremourning/vimspector/103#-LzlW3Oj3Y8P5Vo2iTfn:-LzlW3Oj3Y8P5Vo2iTfo:bzkwk9) ([raw file](https://github.com/puremourning/vimspector/blob/bd794526b71807092ff9357914d235a370bfa912/.gitignore#L18)):_
> 
> > ```gitignore
> > *.vimspector.log
> > support/test/csharp/*.exe*
> > .neomake.log
> > ```
> 
> not sure why this is needed ?
This is smth that my vim setup genetates as an output of an async build command, and, as it is a machine-generated thin (and also irrelevant) i chose to ignore it.
> 
> _[install_gadget.py, line 514 at r1](https://reviewable.io/reviews/puremourning/vimspector/103#-LzlWG6K40VUr9_sVaXp:-LzlWG6K40VUr9_sVaXq:b-u6xxg8) ([raw file](https://github.com/puremourning/vimspector/blob/bd794526b71807092ff9357914d235a370bfa912/install_gadget.py#L514)):_
> 
> > ```python
> >   if not sslcheck:
> >     kwargs = {"context":  ssl._create_unverified_context()}
> > ```
> 
> 1 space after `{` and before `}`
Done;
> 
> Also, i'm not sure that we want to be using private methods of `ssl`.
> 
> Better?
> 
> ```python
>   context = ssl.create_default_context()
>   context.check_hostname = False
>   context.verify_mode = ssl.CERT_NONE
> ```
> 
> flagrantly stolen from https://stackoverflow.com/questions/19268548/python-ignore-certificate-validation-urllib2/28048260#28048260
I have been there:) I just discovered that there was a shorter way, but yes, I agree, using private methods is probably should have avoided. Fixed;
> 
> _[install_gadget.py, line 672 at r1](https://reviewable.io/reviews/puremourning/vimspector/103#-LzlWfeC8smIUi0cvRF6:-LzlWfeC8smIUi0cvRF7:bvfrjtu) ([raw file](https://github.com/puremourning/vimspector/blob/bd794526b71807092ff9357914d235a370bfa912/install_gadget.py#L672)):_
> 
> > ```python
> > parser.add_argument(
> >     "--no-ssl",
> > ```
> 
> prefer to call this `--no-check-certificate` (i think that's actually what `wget` uses) as i think that's clearer (we're not disabling SSL here, we're just ignoring invalid or untrusted certs)
Done;
> 
> _[install_gadget.py, line 675 at r1](https://reviewable.io/reviews/puremourning/vimspector/103#-LzlWOnfDEPtc7F5G8MT:-LzlWOnfDEPtc7F5G8MU:bhyt480) ([raw file](https://github.com/puremourning/vimspector/blob/bd794526b71807092ff9357914d235a370bfa912/install_gadget.py#L675)):_
> 
> > ```python
> >     action = "store_true",
> >     help = "Switch the gadget download ssl certificate verification"
> >     "off (use only in you have a self-signed certificate in your chain)"
> > ```
> 
> missing space in help between `verification` and `off`.
> 
> Also align `"off` with `"Switch` and put the `) on thesame line like the others.
> 
> Tbh, i think the help should be "Do not verify SSL certificates for file downloads."
Done;

BlueDrink9:
Better yet @axrt, put it in your global .gitignore (usually ~/.gitignore). That's where editor-specific files should go. The project .gitignore is for build-files that will be generated by this specific project only, for all users.

axrt:
> Better yet @axrt, put it in your global .gitignore (usually ~/.gitignore). That's where editor-specific files should go. The project .gitignore is for build-files that will be generated by this specific project only, for all users.

Good point. Will do,
cheers!

Issue #97, opened 2020-01-27
[Feature Request]: Ability to programmatically set the DAP communication port.


dansomething:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to take the communication port number result of the `vscode.java.startDebugSession` command and pass it to vimspector when starting it.

https://github.com/puremourning/vimspector/issues/3#issuecomment-578966649

**Describe the solution you'd like**
I think what I'm looking for is a function in the vimspector public API which takes the DAP communication port as an argument. It would be callable from viml.

**Describe alternatives you've considered**
I haven't considered any alternatives though I'm open to anything to streamline the process of using a random DAP communication port and getting that information seamlessly passed to vimspector.

**Additional context**
See this ticket https://github.com/puremourning/vimspector/issues/3#issuecomment-578968090 for additional context.

puremourning:
Looking at it, it's possible without code changes:

* In the adapter config, change `"port": "ask"` to `"port": "${AdapterPort}"`
* `:call vimspector#LaunchWithSettings( { "AdapterPort": the_port } )`

A quick test suggests that works.

puremourning:
I admit, of course, that `vimspector#LaunchWithSettings` is entirely undocumented.

For the record, it accepts a dictionary mapping [variables](https://puremourning.github.io/vimspector/configuration.html#replacements-and-variables) to values.

There's also a _special_ key `configuration` which can be used to launch with a specific debug configuration, e.g. `call vimspector#LaunchWithSettings( #{ configuration: 'Java Launch' } )` will skip asking you which debug config to use and use that one. I [use this](https://github.com/puremourning/.vim-mac/blob/master/ftplugin/python.vim#L106) for running tests under the debugger

dansomething:
Ok, I'll give it a shot. In the long term would you prefer `vimspector#LaunchWithSettings` not be treated as a public api?

puremourning:
Oh it should be public/documented. That’s why I’m leaving this open.

dansomething:
I've confirmed it works exactly as described.

PR #107, opened 2020-01-31
Improve workflow for launching vimspector from scripts

puremourning:
This allows things like PID to connect to to be specified directly in the launch options, e.g. from `LaunchWithSettings` and/or by using a `variables` block and running a `shell` command.

Include an example of how you might make a command line script to launch directly into vimspector with `vim -S script.vim`.

Loosely related to #97 

Fixes some issue with #85


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/107)

LAHumphreys:
Thanks! :)

Issue #105, opened 2020-01-30
Unable to debug python with node 13 environment. "Number of constructor arguments..."


BlueDrink9:
**Describe the bug**

Starting to debug a basic python file opens vimspector windows but no debugging starts. Vimspector encounters an error and crashes.

**To Reproduce**

* `vim test.py`
* `iprint('hello')`
* `<f5>`

Vimspector config file:

```json
{
  "configurations": {
    "univisal: Launch": {
      "adapter": "vscode-python",
      "configuration": {
        "name": "<name>: Launch",
        "type": "python",
        "request": "launch",
        "cwd": "${workspaceRoot}",
        "stopOnEntry": true,
        "console": "externalTerminal",
        "debugOptions": [],
        "program": "${file}"
      }
    }
  }
}
```

**Expected behavior**

A debugging session starts

**Actual behaviour**

Vimspector windows open but no debugging starts.

No output except stderr, and that was the same as the contents of the log.

**Environemnt**

* Version of Vimspector: (e.g. output of `git rev-parse HEAD` if cloned or the
  name of the tarball used to install otherwise)

c6c0144f85a79baa0664c1fa145c029820e2b293

* Output of `vim --version` or `nvim --version`

```
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Jan 26 2020 11:45:38)
Included patches: 1-148
Compiled by Arch Linux
Huge version with GTK3 GUI.  Features included (+) or not (-):
+acl               -farsi             -mouse_sysmouse    -tag_old_static
+arabic            +file_in_path      +mouse_urxvt       -tag_any_white
+autocmd           +find_in_path      +mouse_xterm       +tcl/dyn
+autochdir         +float             +multi_byte        +termguicolors
-autoservername    +folding           +multi_lang        +terminal
+balloon_eval      -footer            -mzscheme          +terminfo
+balloon_eval_term +fork()            +netbeans_intg     +termresponse
+browse            +gettext           +num64             +textobjects
++builtin_terms    -hangul_input      +packages          +textprop
+byte_offset       +iconv             +path_extra        +timers
+channel           +insert_expand     +perl/dyn          +title
+cindent           +job               +persistent_undo   +toolbar
+clientserver      +jumplist          +popupwin          +user_commands
+clipboard         +keymap            +postscript        +vartabs
+cmdline_compl     +lambda            +printer           +vertsplit
+cmdline_hist      +langmap           +profile           +virtualedit
[...]

PR #110, opened 2020-02-02
Support setting environment in neovim

puremourning:
Implement the 'env' option properly(!?) for neovim. Copy and update the _global_ environment when calling `jobstart` and `termopen`.

This makes `debugpy` work in neovim (and possibly some other stuff).

Fun times.

cc @oblitum @BlueDrink9

references #105

### DIFF
--- a/autoload/vimspector/internal/neoterm.vim
+++ b/autoload/vimspector/internal/neoterm.vim
@@ -25,6 +25,27 @@ set cpoptions&vim
 " FIXME: Tidy this map when buffers are closed ?
 let s:buffer_to_id = {}
 
+function! vimspector#internal#neoterm#PrepareEnvironment( env ) abort
+  let old_env = {}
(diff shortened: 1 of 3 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/110)

autoload/vimspector/internal/neoterm.vim
```diff
@@ -25,6 +25,27 @@ set cpoptions&vim
 " FIXME: Tidy this map when buffers are closed ?
 let s:buffer_to_id = {}
 
+function! vimspector#internal#neoterm#PrepareEnvironment( env ) abort
+  let old_env = {}
+
+  let new_env = copy( environ() )
```
puremourning: The copy is not required here, it is wasteful

PR #115, opened 2020-02-08
Switch to debugpy rather than vscode-python

### LINKED ISSUE #109: [Feature Request]: Switch to debugpy rather than vscode-python
See #108  and #105 and a number of other complaints - we can't really rely on the way we're using vscode-python's typescript debug adapter anymore.

* [ ] : Find a way to install `debugpy` using `install_gadget.py`
* [ ] : Make `install_gadget.py` write a compatibility adapter config (should be trivial)
* [ ] : Ensure that the regression tests still pass

puremourning:
Thanks to @int19h for all the help, we're now forcing migration to `debugpy`; the old adapter is still there (for now) but requires specifying `--force-enable-python.legacy` to `install.py`. Now `--enable-python` will locally build `debugpy` in the gadget directory (rather than relying on pip or any of that nonesense). 

Fixes #109

### DIFF
--- a/.vintrc.yml
+++ b/.vintrc.yml
@@ -1,0 +1,3 @@
+cmdargs:
+  color: true
+  severity: style_problem
--- a/.ycm_extra_conf.py
+++ b/.ycm_extra_conf.py
@@ -38,6 +38,13 @@ def Settings( **kwargs ):
       }
     }
 
+  if kwargs[ 'language' ] == 'python':
+    return {
+      'sys_path': [
+        p.join( PATH_TO_THIS_DIR, 'python3' )
+      ]
+    }
+
   if IgnoreExtraConf:
     raise IgnoreExtraConf()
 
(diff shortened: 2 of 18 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/115)

int19h:
@puremourning A heads up that we're also refactoring / cleaning up both the API and the CLI in debugpy (the excuse being that the rename broke backwards compatibility on both counts anyway). You can see what changed here:

https://github.com/microsoft/debugpy/commit/77654f6ff61bf3413a993b5f91b9fe163233793b

This is in debugpy master now, but we haven't pushed a new release to PyPI yet. 

It doesn't look like this should affect you directly, since nothing changed for `debugpy.adapter`. But there might be some sample code that needs updating, if you had `enable_attach()` etc anywhere.

puremourning:
Thanks for the heads up!

PR #116, opened 2020-02-09
Support per-filetype global launch configurations, and support trivial custom gadget installation

puremourning:
This adds the ability to:

* Define filetype-specific configs in the vimpector directory, this will allow projects to be debugged _without_ a `.ycm_extra_conf.py`
* Allow custom adapters to be installed using `install_gadget.py` (with limitations). This will allow users to just drop in a json file and make that language work. 

TODO:

* [ ] : Need a more powerful mechanism to set up the adapter after download (e.g. a 'do' key which takes a script or something)
* [ ] : Specify default configurations in the install meta


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/116)

puremourning:
Let's split this into 2 - filetype config dirs and custom installer stuff, as the latter is much tricker to do well and the former is quite useful, allowing zero-per-project config by dropping filetype configs in the config dir once.

Issue #40, opened 2019-07-04
[Feature Request]: Zero config for trivial projects


puremourning:
**Is your feature request related to a problem? Please describe.**
Configuring vimspector for the project isn't easy if you don't know the options for a given adapter. The READMe has some examples, but that's not easily accessible. VScode provides semantic completion (using json language server and son schema from package.json), but vimspector requires you just know its config or read the instructions n the website.

**Describe the solution you'd like**
Ideally, read package.json and integrate with YCM's json language server support (which doesn't exist) and its snippet completions (which also don't exist). 

But for starters we could just have some UltiSnips snippets which cover the supported languages

Out of scope: snippets for snippet systems I don't use, support for other completion systems I don't use.

**Describe alternatives you've considered**
As above - maybe parse the package.json if it exists (e.g. for vscode-cpptools, vscode-python)

puremourning:
This is still useful, but with adding the json schema, editors can take advantage of that, including at least my own fork of ycm.

tinmarino:
Is it acceptable to write a fallback `GetDefaultConfig` to `PathToConfigFile` if no config file is found: 

https://github.com/puremourning/vimspector/blob/68c47edabb627705e6c8beff1909f63a5041e6cd/python3/vimspector/debug_session.py#L79-L89

If current filetype is python => returns the formated string (replacing current_file AND current_file_pwd)

```json
{
  "configurations": {
    "simple_python: launch": {
      "adapter": "vscode-python",
      "configuration": {
        "name": "Python: Launch current file",
        "type": "python",
        "request": "launch",
        "cwd": "{current_file_pwd}",
        "stopOnEntry": true,
        "console": "externalTerminal",
        "debugOptions": [],
        "program": "{curren_file}"
      }
    }
  }
}
```

That way, for a short standalone file, the user do not have to create a config file, just press `F5` ?

What do you think ? I'm coding that and make a PR ?

puremourning:
I think it's simpler to just put a `.vimspector.json` in your home directory.

Alternatively, if that's not great, then we could perhaps have something like `${gadgetDir}/.conf.d/<filetype>.defaults.json` which contains default debug configs for `<filetype>`. 

The [existing variable replacements](https://puremourning.github.io/vimspector/configuration.html#predefined-variables), like `${workspaceRoot}` would have to be adjusted when using configs from that locaiton.

I wouldn't merge a PR which includes adapter-specific configuraiton in the core code, no.

[...]

PR #120, opened 2020-02-15
Support for scala

danilbykov:
I would like to add couple functions to support debugging of scala applications. It will work a bit different than current workflow.

[Metals](https://scalameta.org/metals/) is language server for scala. But besides usual possibilities of language server, Metals can start Debug Adapter for any runnable application or test from project. So using CodeLens from Language Server Protocol, the user can launch application (or test) and start Debug Adapter. It means that gadgets and file `.vimspector.json` are not needed for scala developers. Metals can start Debug Adapter for free. Once debug server is launched Metals returns port where server is running. So vimspector just needs to connect to this port.

In my pull request I add new function `Connect` which will be called from [coc-metals](https://github.com/scalameta/coc-metals) extension. So user will be able to debug application using CodeLens from Language Server Protocol without gadget or file `.vimspector.json`. I already [talked](https://github.com/scalameta/coc-metals/issues/80) with author of coc-metals and he agreed to add this integration in coc-metals.

### DIFF
--- a/autoload/vimspector.vim
+++ b/autoload/vimspector.vim
@@ -30,6 +30,10 @@ function! vimspector#LaunchWithSettings( settings ) abort
   py3 _vimspector_session.Start( launch_variables = vim.eval( 'a:settings' ) )
 endfunction
 
+function! vimspector#Connect( port, configuration ) abort
+  py3 _vimspector_session.Connect ( port = vim.eval( 'a:port' ), configuration = vim.eval( 'a:configuration' ) )
+endfunction
+
 function! vimspector#Reset() abort
   py3 _vimspector_session.Reset()
 endfunction
--- a/python3/vimspector/debug_session.py
+++ b/python3/vimspector/debug_session.py
@@ -77,6 +77,7 @@ class DebugSession( object ):
                        launch_variables )
     self._configuration = None
     self._adapter = None
+    self._supportReconnect = True
 
     current_file = utils.GetBufferFilepath( vim.current.buffer )
 
@@ -246,12 +247,24 @@ class DebugSession( object ):
 
     start()
 
+  def Connect( self, port, configuration ):
+    self._logger.info( 'Connecting to %s using %s', port,
+                       json.dumps( configuration ) )
+    self._supportReconnect = False
+    self._StartWithConfiguration(
+      configuration,
+      { 'port': int( port ) }
+    )
+
   def Restart( self ):
     # TODO: There is a restart message but isn't always supported.
     # FIXME: For some reason this doesn't work when run from the WinBar. It just
     # beeps and doesn't display the config selector. One option is to just not
     # display the selector and restart with the same opitons.
-    if not self._configuration or not self._adapter:
+    if not self._supportReconnect:
+      utils.UserMessage( 'Restart is not supported in current mode.' )
+      return
+    elif not self._configuration or not self._adapter:
       return self.Start()
 
     self._StartWithConfiguration( self._configuration, self._adapter )


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/120)

danilbykov:
> However, I think this additional API is not necessary, you can already do this with the existing LaunchWithSettings.

Yes, I can start debugging using function `LaunchWithSettings`. Here is steps which I need:
1. Create dummy `.vimspector.json` (in each scala project which I want to debug)
```
{
  "configurations": {
    "Test": {
      "adapter": {
        "port": "ask",
        "variables": {
        }
      },
      "configuration": {
        "name": "abc",
        "request": "launch"
      },
      "breakpoints": {
        "exception": {
          "caught": "N",
          "uncaught": "N"
        }
      }
    }
  }
}
```
This file doesn't contain any useful information.
2. Then I need call `vimspector#LaunchWithSettings({configuration: 'Test', test: 'abc'})`
3. Take port from Metals where debug server is started and enter this port for vimspector (manually). Metals always starts debug adapter on random free port.

Can I simplify this process?

With new function `Connect` I have following steps:
1. Metals provides CodeLens for all `main` functions and all tests. These CodeLens allows to start Debug Adapter for test or application automatically. So user runs CodeLens using some hotkey.
2. Debug Adapter is started.
3. [coc-metals](https://github.com/scalameta/coc-metals) extension gets port where Debug server is started and calls function `Connect` of vimspector.

As result user doesn't need dummy `vimspector.json`, call function `LaunchWithSettings`, enter port.

> I'm unconvinced that there are and never will be launch configuration options for the metals server

Metals server is Language Server (not Debug Adapater). This server is started by coc-metals and coc-metals has different configuration options. Maybe you mean Debug Adapter server. I can imagine that Debug Adapter needs environment and program arguments. This is still open feature request for Metals:
https://github.com/scalameta/metals-feature-requests/issues/63
But I can not understand how file `.vimspector.json` can help us. Metals doesn't know anything about this file and can not take these arguments from this file.

> How is the debug adapter initiated in metals ?

Client (vim) sends this command to Metals:
```
{
  "command": "debug-adapter-start",
  "arguments": [
    {
      "targets": [
        {
          "uri": "file:/home/dbykov/reps/test/experiments/?id\u003droot"
        }
      ],
      "dataKind": "scala-main-class",
      "data": {
        "class": "monix.HotObservable",
        "arguments": [],
        "jvmOptions": []
      }
    }
  ]
}
```

puremourning:
> Can I simplify this process?

yes.

> 3. Take port from Metals where debug server is started and enter this port for vimspector (manually). 

You don't have to do it manually, you can use a replacement var in the `.vimspector.json`:

`"port": "${port}"`

And then pass the port in the dict passed to `LaunchWithSettings`

> But I can not understand how file .vimspector.json can help us. Metals doesn't know anything about this file and can not take these arguments from this file.

With the language-specific dirs proposed in #40 we would just put the trivial config file in `/path/to/vimspector/configurations/scala/metals.json` then you don't need a `.vimspector.json` in the project root.

To explain, I don't want to accept PRs which improve complexity of vimspector and only improve the workflow for one use-case. What I'm suggesting is to improve the workflow generally, which also improves this one to the same degree.

While I might be open to a `Launch`-like API which accepts a port to connect to, I just don't think there are enough debug adapters in the world which don't require configuration to be specified as part of the `initialize` exchange to mean that we can do away with launch configuration. I'm also not convinced that the debug adapter within metals has no such arguments. Do you have a link to the code which proves it ? 

I mean, as an example, 'request' is mandatory. How would "attach" requests work ?

danilbykov:
Thanks for hint with `${port}`, it works. So I just need "empty" `.vimspector` file.

I'm closing this PR.

puremourning:
Cool. 

I think I will merge the "0-config" pr if the tests are green. then you can put the file 1x in your vimspector install and all will just work without the 'empty' vimsecptor file per-project.

Thanks for the contribution and the discussion though!

PR #123, opened 2020-02-16
Set some window options to their deafults

puremourning:
This makes the UI feel more like a UI.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/123)

PR #124, opened 2020-02-16
List splat

puremourning:
Support splat for variable expansion in lists.

The motivating example for this is to allow shell arguments to be specified like:

```json
"program" : "/path/to/a/script",
"args": [ "arg1", "*${args}", "arg2" ]
```

or some similar. We want to ask the user for some command line arguments, but:

* (optionally) we want to supply our own args before or after them (or both)
* we _must_ supply the args as a list

### DIFF
--- a/python3/vimspector/utils.py
+++ b/python3/vimspector/utils.py
@@ -21,6 +21,9 @@ import vim
 import json
 import string
 import functools
+import subprocess
+import shlex
+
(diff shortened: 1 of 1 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/124)

PR #136, opened 2020-03-21
Support c-style comments in JSON

### LINKED ISSUE #135: [Feature Request]: Allow comments in .vimspector.json
`.vimspector.json` is similar to VSCode's `tasks.json` or `launch.json` but unfortunately I cannot put any comments there.

It would be great to switch to [JSON with Comments (jsonc)](https://code.visualstudio.com/docs/languages/json) for vimspector config files.

puremourning:
Debate rages about whether JSON should have comments. The specification
says it shouldn't but the author of JSON suggested that if you want to
use JSON for configuration, then pipe it through jsmin before parsing.

So that's what we do, using a tiny JSON minifier from
https://github.com/getify/JSON.minify/tree/python

Closes #135

### DIFF
--- a/README.md
+++ b/README.md
@@ -609,6 +609,11 @@ the Getting Started section of the [Vimspector website][website].
 For full explanation, including how to use variables, substitutions and how to
 specify exception breakpoints, see [the docs][vimspector-ref].
 
+The JSON configuration file allows C-style comments:
+
+* `// comment to end of line ...`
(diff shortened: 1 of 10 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/136)

noscript:
Thanks a lot!

PR #140, opened 2020-03-26
Add a splash popup while starting up

puremourning:
Sometimes it can take quite a while to start up and initialise the debug
adapter. So we use popup/float to display the status as we start up and
shut down.

This increases minimum Vim version to 8.2, but that's been out for ages
now and I intend to agressively require latest/later vim/neovim
versions.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/140)

PR #142, opened 2020-03-27
Tweaks

puremourning:
* Add `autoselect` option to debug configs. Setting it to `false` means that this option won't be automatically picked when hitting `<F5>`. Useful for configs in the shared `configurations` directory.
* Ignore the `loadedSource` event. it's not clear what it's for, but some adapters spam it out and it's noisy to see the errors
* Don't terminate the deubgee when detaching. vscode-cpptools does this irrespective and it's really annoying.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/142)

PR #148, opened 2020-04-03
Remove workaround for broken cpptools

puremourning:
Thanks to @neiljohari  for updating cpptools. This change removes all of the workarounds we had in place.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/148)

PR #149, opened 2020-04-04
debugpy updates

dianlujitao:
For some reason `additionalProperties` has higher priority over `properties` inside the conditional statement thus I removed it as a workaround:\

### DIFF
--- a/docs/schema/debugpy.schema.json
+++ b/docs/schema/debugpy.schema.json
@@ -1,0 +1,289 @@
+{
+  "$id": "https://puremourning.github.io/vimspector/schema/debugpy.schema.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "oneOf": [
+    {
+      "properties": {
+        "request": {
+          "const": "launch"
+        },
+        "module": {
+          "type": "string",
+          "description": "Name of the module to be debugged.",
+          "default": ""
+        },
+        "program": {
+          "type": "string",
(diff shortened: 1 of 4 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/149)

dianlujitao:
Yes I totally agree that maintaining the documents would be a big mess. Unquestionably pointing to upstream maintained ones would be the best solution, MS didn't provide such things. They don't release the schemas with debuggers but vscode plugins.

For c/cpp the schema is [here](https://github.com/microsoft/vscode-cpptools/blob/ac2398d98730da27de65382b86258f0bf0878c31/Extension/package.json#L638), but further postprocessing step is applied to descriptions for i18n, which is not natively supported by json schema (to my knowledge).

In case we wanna document these debugger-specific options, we have no choice but to maintain by ourselves. One possible solution might be automating the process to extract the schema from package.json with scripts, but given the diversity of different vscode plugins, per-language adaption is still required.

I'm pretty unsure what should be done, the decision is up to you.

puremourning:
I think (for now at least) I would like to split the update of debugpy out (and ship that). Then we can discuss the options about the json schema ? Would you be willing to do that?

dianlujitao:
ofc, feel free to do what you want:D

puremourning:
Thanks, I stole your change and hid it in #151 ; I'll leave this open for the discussion.

PR #150, opened 2020-04-05
Add troubleshooting guide

puremourning:
Also make sure that the gadget installer and other docs are included in
the tarball installation.

cc @dboeckenhoff who pointed out that install_gadget.py was missing from the tarball installation!


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/150)

PR #153, opened 2020-04-25
Add basic support for conditional breakpoints

### LINKED ISSUE #152: [Feature Request]: Conditional breakpoints
Currently three types of breakpoints are supported (function, line and exception breakpoints). This has already been extremely helpful when working on larger projects. However, often times I find myself wanting conditional breakpoints in order to only pause the program when a very specific situation is occurring.

I believe the protocol supports conditional breakpoints and projects such as [emacs dap](https://github.com/emacs-lsp/dap-mode#debugger-commands) supports adding them. 

Currently I find myself running lldb directly or switching to another editor such as emacs when dealing with conditional breakpoints and then switching back to vim when I want to keep working, but it would be great to have everything in one place.

Thanks again for all your hard work on this project.

puremourning:
This is the minimal required for a user to use conditional breakpoint -
we add an options dict to each breakpoint (line and function) and allow
the condition to be supplied. We add a plug mapping and a default
shortcut (<leader><F9>) to add one where we ask the user to enter the
condition and hit expression. This isn't great but it works.

We don't check the capabilities, so they would just be ignored if used
on a server that doesn't support them. We also ask for a hit expression
which most users won't understand so this isn't ideal either.

No tests, and no documentation yet.

Added a new sign type too, so you can see where a breakpoint is conditional.

I actually use this:

```
" vimspector
let g:vimspector_enable_mappings = 'HUMAN'

sign define vimspectorBP text=🔴         texthl=Normal
sign define vimspectorBPCond text=⭕     texthl=Normal
sign define vimspectorBPDisabled text=🔵 texthl=Normal
sign define vimspectorPC text=🔶         texthl=Normal
```

Fixes #152

### DIFF
--- a/README.md
+++ b/README.md
@@ -458,6 +458,7 @@ features to set your own mappings. To that end, Vimspector defines the following
 * `<Plug>VimspectorRestart`
 * `<Plug>VimspectorPause`
 * `<Plug>VimspectorToggleBreakpoint`
+* `<Plug>VimspectorToggleConditionalBreakpoint`
 * `<Plug>VimspectorAddFunctionBreakpoint`
 * `<Plug>VimspectorStepOver`
 * `<Plug>VimspectorStepInto`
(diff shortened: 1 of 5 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/153)

adam-r-kowalski:
I have tested this branch locally and can confirm that conditional breakpoints work both in vim8 and neovim!

PR #160, opened 2020-05-09
Allow moving the gadget/configuration dirs to arbitrary location

puremourning:
This is handy for running the tests in a clean way (e.g. if you have a configurations directory or custom gadgets, etc.)

It could also be useful in multi-user environments and/or (unfortunately) for third-party packaging (which I _strongly_ discourage)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/160)

PR #164, opened 2020-05-12
Make sure we can still request threads if one threads request fails

### LINKED ISSUE #163: Vimspector + netcoredbg is not functional.
Hi. Recently I decided to give a c# + netcore + vim  another try and discovered vimspector. Unfortunately when I tried a simple console app debugging it failed. Vimspector screen would show up and I could see the code window but Stacktrace, Watches and other were empty. Upon looking in the logs I discovered that DAP `threads ` command fails with 0x80131302 error code (Process not synchronized).  

The main reason of this behavior seems to be netcoredbg which fails to perform `threads ` command when not in stopped/breakpoint state. After commenting `threads` command in vimspector everywhere except in `stopped` event handler vimspector started to work (or seemed so) which is a temporary solution apparently. 

Netcoredbg would fail on any `threads` command after `threads` event because of that as well. I am aware that according to DAP specification adapter should return dummy threads in if it is impossible to get the current threads state (after startup phase). 

While the root cause of this issue is `netcoredbg` DAP implementation and bug must be addressed on their side I still think that failed `threads` command should be treated correctly in vimspector and not make it look frozen (must be due `_requesting_threads `variable not resetting? didnt have time to go deep.) 

System specs:
.netcore 3.0/3.1
neovim 0.4.3/vim 8.1
Windows WSL2 or Mac OS 10.15  (repro in both)

Log file: 
[vimspector.log](https://github.com/puremourning/vimspector/files/4614212/vimspector.log)

puremourning:
Fixes #163

### DIFF
--- a/python3/vimspector/stack_trace.py
+++ b/python3/vimspector/stack_trace.py
@@ -118,10 +118,14 @@ class StackTraceView( object ):
 
       self._DrawThreads()
 
+    def failure_handler( reason, msg ):
+      # Make sure we request them again if the request fails
+      self._requesting_threads = False
+
     self._requesting_threads = True
     self._connection.DoRequest( consume_threads, {
       'command': 'threads',
-    } )
+    }, failure_handler )
 
   def _DrawThreads( self ):
     self._line_to_frame.clear()


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/164)

Issue #11, opened 2019-04-27
Remote debugging is not documented


puremourning:
Vimspector supports some fairly powerful remote debugging features, as required by the author. These options are not documented and there is no tutorial.

puremourning:
Tutorial will be added to the website: https://puremourning.github.io/vimspector-web

puremourning:
It's sort of documented in the reference guide: 

```
   "adapter-remote": {
      "properties": {
        "remote": {
          "type": "object",
          "required": [ "host" ],
          "description": "Configures how Vimspector will marshal remote debugging requests. When remote debugging, Vimspector will ssh to 'account'@'host' and run 'pidCommand', 'attachCommands', 'runCommands', etc. based on the 'remote-command' option in the debug configuration. If 'remote-command' is 'launch', it runs 'runCommand(s)', otherwise (it's 'attach') vimspector runs 'pidCommand', followed by 'attachCommand(s)'.Then it starts up the debug adapter with the debug configuration as normal. Usually this is configured with an 'attach' request (whether we remotely 'launched' or not). Once the initialization exchange is complete, Vimspector runs the optional 'initCompleteCommand' which can be used to force the application to break, e.g. by sending it SIGINT. This is required on some platforms which have buggy gdbservers (for example)",
``

gdshen:
After read this issue and documentation, I still could not figure out how to configure remote debug.

Could you provide a full configuration for remote debug using gdbserver? And if the target device does not support ssh connection, can we connect to gdbserver directly without ssh using vimspector?

Thanks.

puremourning:
just use this: https://code.visualstudio.com/docs/cpp/launch-json-reference#_midebuggerserveraddress; set it to what you would put in a gdb `target remote` command.

you'll need to get gdbserver running on the device manually of course. vimspector can only do it for you if SSH is used.

PR #166, opened 2020-05-15
Debugpy remote debugging and install issues

puremourning:
Improves install_gadget.py for shared environments:

- move directories out of the way if they dan't be deleted (e.g. NFS)
- adds `--update-gadget-config` to allow incremental installs of gadgets rather than having to do all of them every time (or manually with `--no-gadget-config`)

Improves debugpy support:

- implements the "multi-session" mode for DAP. We basically supported this already for local DA (java). We just need to specify a host too
- documents remote debugging features (#11 )
- adds an attach test for python and migrates to debugpy


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/166)

PR #167, opened 2020-05-15
Fix incorrect naming of `.viminspector.json`

esetnik:
I got tripped up getting started because the documentation incorrectly uses `vimspector.json` and I was getting an error about no debug configurations found.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/167)

PR #170, opened 2020-05-17
add docker exec as an attach command

awwalker:
adds a second method of connection during an attach session via `docker exec -it`


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/170)

olimorris:
@awwalker do you have an example dockerfile you used when creating this? I'm having issues trying to use the example config you supplied, locally.

Issue #141, opened 2020-03-27
[Feature Request] Multiple concurrent sessions (to debug multiprocessing python programs)


MoetaYuko:
**Describe the bug**
Can't debug multiprocessing python programs

**To Reproduce**
MWE

```python
import time
import multiprocessing as mp


def f():
    while True:
        print("f")
        time.sleep(0.1)


print("main")
p = mp.Process(target=f)
p.start()
p.join()
```

List of steps to reproduce

1. Set a breakpoint in `f()`
2. Start debugging

Vimspector config file:

```
{
  "$schema": "https://puremourning.github.io/vimspector/schema/vimspector.schema.json#",
  "configurations": {
    "Current file: Launch": {
      "adapter": "debugpy",
      "configuration": {
        "name": "Current file: Launch",
        "request": "launch",
        "console": "externalTerminal",
        "program": "${file}"
      }
    }
  }
}
```

**Expected behavior**
Stop at the breakpoint.

**Actual behaviour**
Stop at `p.join()` and stuck there.

Please include:
* Vimspector log (~/.vimspector.log)

http://fars.ee/l00b

* Output from any or all UI diagnostic tabs (Server, etc.)

```
Unhandled event: debugpyAttach
```

**Environemnt**

NOTE: NeoVim is supported only on a best-effort basis. Please check the README
for limitations of neovim. Don't be offended if I ask you to reproduce issues in
Vim.

NOTE: Windows is not supported. There is a branch with windows support, which
you can contribute to, rather than opening an issue.

* Version of Vimspector: 87e80db94d0623043d591f10d706cc50d1acc4f0

* Output of `vim --version` or `nvim --version`

```
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Mar  1 2020 17:21:03)
Included patches: 1-343
Compiled by Arch Linux
Huge version without GUI.  Features included (+) or not (-):
+acl               -farsi             -mouse_sysmouse    -tag_old_static
+arabic            +file_in_path      +mouse_urxvt       -tag_any_white
+autocmd           +find_in_path      +mouse_xterm       +tcl/dyn
+autochdir         +float             +multi_byte        +termguicolors
-autoservername    +folding           +multi_lang        +terminal
-balloon_eval      -footer            -mzscheme          +terminfo
+balloon_eval_term +fork()            +netbeans_intg     +termresponse
-browse            +gettext           +num64             +textobjects
++builtin_terms    -hangul_input      +packages          +textprop
[...]

PR #172, opened 2020-05-19
Disable subprocess debugging by default in debugpy

puremourning:
As mentioned here: https://github.com/puremourning/vimspector/issues/141#issuecomment-605289072


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/172)

PR #173, opened 2020-05-20
Document that Vim have to be built with 'terminal' feature support

ardazishvili:
Dear Ben,

Thanks for a good tool! I just want to mention in the README that Vimspector is using 'terminal' feature of the Vim, so Vim should be built with 'terminal' support.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/173)

puremourning:
I made a change to the doc for this. Thanks again for pointing this out!

PR #179, opened 2020-06-02
fix some exceptions not handled

chinnkarahoi:
I duplicate vimspector.Variables in a new window/tab which cannot be expanded

### DIFF
--- a/python3/vimspector/utils.py
+++ b/python3/vimspector/utils.py
@@ -171,8 +171,11 @@ def RestoreCurrentWindow():
   try:
     yield
   finally:
-    vim.current.tabpage = old_tabpage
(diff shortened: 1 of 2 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/179)

puremourning:
Thanks for sending a PR! Really appreciate you taking the time to contribute to vimspector!

I'll try and look at this later, but it seems to be addressing multiple things.

could we try and keep PRs to one thing at a time, describing the problem, solution and how it is tested/reproduced ?

chinnkarahoi:
Thanks for replying. I will close this PR and create multiple PRs for addressing things

PR #181, opened 2020-06-04
ExpandVariable for the buffer of vimspector.variables rather than the window

chinnkarahoi:
Launch vimspcetor and focus on vimspector.variables window, type `:split<cr>`, pressing enter at any line starting with `+` has no effect in the spawned window


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/181)

puremourning:
Implemented in #197 

Going to close this in favour of that.

PR #183, opened 2020-06-05
Implement `Don't trigger autocommands when shifting windows`

chinnkarahoi:


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/183)

puremourning:
I guess this is abandoned. and tbh I'm not sure that it's all that necessary. the original comment was due to some perceived slowness, but that turned out to be the adapter doing a lot of logging. 
I'm going to close this for now.

PR #190, opened 2020-07-03
Support running a command and connecting to a port

### LINKED ISSUE #145: [Feature Request]: CodeLLDB adapter support

I've been having issues with vscode-cpptools for a while (https://github.com/microsoft/vscode-cpptools/issues/3829#issuecomment-606314684). The recent comments suggest it may have been patched but I've been sticking with CodeLLDB since it works.

It would be nice if install_gadgets could add support for CodeLLDB. I am familiar enough with how vimspector works to add a new gadget and send in a PR, though I don't mind playing around with it for a bit given some pointers.

I've tried using vscode-cpptools and lldb-vscode but have issues with input redirection.

puremourning:
CodeLLDB can be launched outside of VScode, but it still listens on a port rather than uses stdio.

Support debug adapter configuration that launches a command then connects to a port opened by that command.

This makes it possible to use CodeLLDB (aka vscode-lldb) with Vimspector. It works well and is probably the best supported server for rust debugging.

fixes #145 

Should really update the docs for this but maybe another PR can add an install.py option to just download the "platform" bundles from the lldb-vscode repo and sets up the default adapter config.

For now, included a trivial example in support/test/rust. However, there is a slight issues that "Restart" doesn't work. I _think_ because of the port timeouts.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/190)

puremourning:
cc @GopherJ

GopherJ:
Hi @puremourning Great thanks on making this PR. I manually added your changes on the branch `codelldb`, when I launch vimspector I got:

image

puremourning:
>  I manually added your changes on the branch

what do you mean "manually" ?

GopherJ:
I'm using vimspector's tarball installation so actually I didn't find a good way of pulling your changes. I edited manually the files, I'm still verifying if I made an error

puremourning:
Oh, hah. sure. you probably missed this https://github.com/puremourning/vimspector/pull/190/files#diff-63bed83e2eb76fb956ba5c86970d12abR779

GopherJ:
@puremourning yes exactly there~

GopherJ:
@puremourning Awesome just tested it works like charm. CodeLLDB's rust support is much better than vscode-cpptools

image

Thanks @puremourning for your awesome work and help! Will explorer more this weekend!

GopherJ:
@puremourning Hi, wondering if `hover` works for you? I used F6 to set pause and then moved my cursor on a variable but I cannot see value displayed.

I'm using vim 8.2

puremourning:
> @puremourning Hi, wondering if `hover` worked for you in rust? I used F6 to set pause and then moved my cursor on a variable but I cannot see value displayed.
> 
> I'm using vim 8.2

Yes it works for me. All mouse event support depends on your terminal emulator supporting it. try with something known to work like xterm

GopherJ:
Have you tested if it works in tmux?  I'm checking if someone configs need to be applied to make it works

puremourning:
> Have you tested if it works in tmux? I'm checking if some configs need to be applied to make it works.
> 
> **EDIT**: It works fine in deepin-terminal or others but not in tmux

https://github.com/tmux/tmux/issues/55

GopherJ:
@puremourning Figured it out.

I need to add:

```
if !has('nvim')
    set ttymouse=xterm2
endif
```
in `.vimrc`

to make it works in tmux


So now the only problem on my side is the hover text, I've seen `^@` for some reason:

image

puremourning:
> So now the only problem on my side is the hover text, I've seen ^@ for some reason:

That's unrelated to this issue, but the `^@` are newlines. The balloon popup probably needs to use balloon_split. 

Feel free to raise a new issue for that with a repro case.

GopherJ:
@puremourning Yes, that's another topic.

FYI, I noticed crashes in neovim 0.4.3 for the same test:

image

I don't use neovim so it's ok...

puremourning:
I haven't implemented the job control for neovim yet. As I have said many times on reddit and to anyone who will listen, supporting neovim doubles my workload...

GopherJ:
@puremourning emm... I certainly know this, but it didn't happen when I use vscode-cpptools, that's why I put it here as the information may be useful

puremourning:
@GopherJ  thanks for testing. i've pushed support for this in neovim.

GopherJ:
@puremourning great, thanks!

PR #192, opened 2020-07-07
Fix shutting down jobs when resetting - use the correct category for …

puremourning:
…CleanUpCommand


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/192)

PR #196, opened 2020-07-09
Automatically pick a local port for CodeLLDB

### LINKED ISSUE #188: [Feature Request]: Remove all breakpoints
Once I'm done with a debugging session, I want to get back to normal coding and not have breakpoints set and forget about my debug session. When I 'Reset' at the end, the breakpoints are still remembered (which seems right to me), but it would be nice to also have a 'Remove all breakpoints' action that lets me truly clear the breakpoints and get back to square one.

The 'Remove all breakpoints' action described above.

I browsed around the docs a bit but I don't think I could figure out how to do something like this with the current featureset. 🤔

puremourning:
Also closes #188

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,6 +307,30 @@ should be):
 ./install_gadget.py --all --disable-tcl
 ```
 
+If you want to just add a new adapter without destroying the exisitng ones, add
+`--update-gadget-config`, as in:
+
+```bash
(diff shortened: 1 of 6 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/196)

PR #198, opened 2020-07-11
UI customisation

### LINKED ISSUE #86: [Feature Request]: Customise UI
There are three things that I would love to be able to configure from my `.vimrc` or `.vimspector.json`. If these options already exist then they haven't been documented in an easy to find way, as I've spent a while looking!

1. Persist a default option for the questions `Break on all Exceptions?` and `Break on all Uncaught Exceptions?`. As right now I have to answer them everytime I launch the debugger. 
2. Configure what windows are displayed. Personally I don't use the StackTrace, Watcher and Variables windows. It would be nice if they could be disabled, to give more room for the other windows. 
3. Configure placement and size of the debug windows. It would be great to define the % splits of each window. Something similar to [Nuake](https://github.com/Lenovsky/nuake) would be awesome. 

My workflow may be a bit different to others, as I work with TypeScript, so I thought you might like to hear about how I would ideally use `vimspector` in case it is totally different to how you would. I'd like to keep it really simple, be browsing and working on code, hit a key to add break points, then another to launch the debugger. 

Rather than use watchers or variable windows, I would love to just have a responsive console (Like the JS developer console in chrome) that I could use to inspect and interact with variables. If it would be possible to have this (and the application output) open as Nuake does this would be excellent. 

This is just my personal wishlist though, I started looking at how to achieve some of these things myself, but python+vimscript is new to me so I got a little scared!

Thanks for making vimspector, it's amazing, and something that has been missing from the Vim-ecosystem for ages! :D <3

### LINKED ISSUE #195: [Feature Request]: launch terminal in program output window
The terminal window makes the code window very small and since there is no way to configure scaling of the windows I need to resize them manually every time I run the debugger

it will be great if it can be a tab in the output window.

option to hide/show windows.
making the program output window smaller can mitigate the issue.

**Additional context**
Screen Shot 2020-07-09 at 13 52 13

### LINKED ISSUE #202: UI is messed up when equalalways is set
**Describe the bug**

The UI is poorly proportioned when equalalways is set. This is the default sadly. 

**To Reproduce**
 Set equalalways. Launch. 

**Expected behavior**

UI is proportioned withcode window largest, output window narrow etc. 

**Actual**

Windows are equal. 

UI code tries to unset it while building but I guess it doesn’t t work.

puremourning:
Would love to hear feedback.

Closes #86 
closes #202 
closes #195


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/198)

PR #205, opened 2020-07-15
Hide splash screen when debugger is stopped

lilwayne1556:
I ran into an issue where the splash screen "Shutting down debug adapter..." would stay on the screen when the debugger was stopped and the screen wasn't reset.

I wasn't able to find a better spot to hide the splash screen and I couldn't see a place where the splash screen would have been hidden elsewhere without utilizing the Reset method


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/205)

PR #182, opened 2020-06-04
fix if variable is None; don't print (<unknown type>)

chinnkarahoi:
go version go1.13.8 linux/amd64
```go
package main

func main() {
	a := 0
	b := 1
	c := 2
	d := map[string]int {
		"0": 0,
		"1": 1,
		"2": 2,
	}
	print(a, b, c, d)
}
```
```json
{
  "configurations": {
    "go": {
      "adapter": "vscode-go",
      "configuration": {
        "program": "$cwd",
        "request": "launch",
        "args": [],
        "environment": [],
        "cwd": "$cwd",
        "mode": "debug",
        "dlvToolPath": "$HOME/go/bin/dlv"
      }
    }
  }
}
```
Add a breakpoint at `print` and launch vimspector. Expand `d` in vimspector.variables and  vimspector throws `AttributeError: 'NoneType' object has no attribute 'get'` resulting in following status  
```go
- Scope: Local
  - a (<unknown type>): 0
  - b (<unknown type>): 1
  - c (<unknown type>): 2
  - d (<unknown type>): <map[string]int> (length: 3)
    - "0" (<unknown type>): 0
```
Type definitions are always <unknown type> when debug golang

PR #206, opened 2020-07-16
A number of Bugfixes

### LINKED ISSUE #203: Number column visible with relative number
**Describe the bug**
Watches, variables, stack trace and terminal show number column when use has relative number set. 

**To Reproduce**
Set relative number. Launch. 

**Expected behavior**
Number col not visible in utility windows. 

**Actual***

Number col visible. 

We set nonu but need to set more.

puremourning:
c6e8e80 (HEAD -> bugfixes) Remove hard-coded terminal width which doesn't work on smaller width screens
a4f3fd9 Disable relativenumber in utility windows; disable textwidth in neovim too
2440a98 (origin/bugfixes) Tidy unknwon types and values
434a6f4 Only expand one cheap scope by default - don't keep others expanded from previous frames
fc7725f Only expand one inexpensive scope; codeLLDB lies about how expensive things are
0938d72 Re-use a window if we can, as it's more efficient; don't wipe out the generated code buffers in case the code window gets used for temporary buffer switch
8d2ea44 Make Restart and Stop work properly (and together)
97a5d73 Ignore venv and test basedir
b3c86bc Use the original java test code

These are a rollup of issues found while testing the ui-custom branch.

Closes #205 
Closes #203 
Closes #182

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ README.md.toc.*
 support/test/csharp/*.exe*
 .neomake.log
 configurations/
+venv/
+test-base/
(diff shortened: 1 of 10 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/206)

PR #208, opened 2020-07-18
Put the cursor at the end of the buffer after evaluations

### LINKED ISSUE #187: Use [VimspectorEval] to check variable but the debug output buffer does not automatically scroll to the bottom line
## Environment 
- ubuntu 20.04
- neovim  v 0.4.3
- c debug program

Hey there, thanks for providing such a powerful plugin.
I like to use the command **VimspectorEval** to output the status of the variable, but in use, I found that each time the **VimspectorEval** command is executed, three sentences are output, which are respectively:
- Evaluating:
- Evaluated:
- **result: (That the value i need)**

For these three sections of output, the last one is actually what I really need, but please forget about it. That's not the point I'm making. The most important thing is that the cursor will not be positioned on the last line after every output of these three sentences, and the status will be as follows

微信截图_20200611221507

Then, I have to manually scroll to the last line again to get the information I really need, which is the return value of the **result**

微信图片_20200611222117

I tried to improve this experience with the following vimscript, but it didn't seem to work
ps: I'm a vim novice, not familiar with vim scripts, thanks for lot!
```vim
function! s:PrintVariable(_val)
  execute 'VimspectorEval '. a:_val
  call feedkeys("G") " jump the last line
endfunction
command! -nargs=1 Check call s:PrintVariable(<f-args>)
```
BTW, I don't think **Evaluating** and **Evaluated** result will seem to matter to most people (from my point), maybe removing it will improve the experience of using this plugin.

puremourning:
Closes #187

### DIFF
--- a/python3/vimspector/output.py
+++ b/python3/vimspector/output.py
@@ -126,20 +126,13 @@ class OutputView( object ):
     self._ShowOutput( category )
 
   def Evaluate( self, frame, expression ):
-    console = self._buffers[ 'Console' ].buf
-    with utils.ModifiableScratchBuffer( console ):
-      utils.AppendToBuffer( console, 'Evaluating: ' + expression )
+    self._Print( 'Console', [ 'Evaluating: ' + expression ] )
 
     def print_result( message ):
-      with utils.ModifiableScratchBuffer( console ):
-        utils.AppendToBuffer( console,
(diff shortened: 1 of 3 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/208)

PR #209, opened 2020-07-19
Simplify and improve gadget installation - add VimspectorInstall and VimspectorUpdate

puremourning:
Add in-vim commands to install and upgrade gadgets.

* [x] Pull out the actual installer code into python modules 
* [x] Switch install_gadget.py to use that
* [x] Add some wrapper to call the script from vim
* [x] Add some way to make the arguments more user-friendly
* [x] Add some command line completion for the arguments
* [x] change the defaults to _update_ rather than overwrite gadget config
* [x] re-use the OutputView to show the running output
* [x] suggest gadgets to install when trying to debug and adapter is not found
* [x] update the documentation to simplify the installation instructions
* [ ] find a way to provide a guided/templated initial configuration
* [x] Add VimspectorUpgrade to read the adapter config and reverse-install the the gadgets

The amount of work required here is why we use a script in the fist place - so much simpler. How much will we gain by doing this vs other roadmap items?

Also Fixes https://github.com/puremourning/vimspector/issues/210


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/209)

puremourning:
It's worth noting that a completely different approach could be taken:

* Create `VimspectorInstall` which just runs `install_gadget.py` with whatever available `python3` 
* Default the arguments as above (update by default)
* Find a way to auto-complete the arguments (maybe run the script with `--list-gadgets` or something)
* Again, re-use existing run-with-log-output vimscript usd by vimspector.

This would be a lot simpler, and achieve all of the same goals, but still relies on the user's python runtime...

PR #212, opened 2020-07-22
Fix windows installer issues - use symlinks not junctions and fix dum…

puremourning:
…b use of PIPE with check_call


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/212)

PR #213, opened 2020-07-22
Add bang versions of install commands to leave the output open

puremourning:


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/213)

PR #215, opened 2020-07-22
Add ability to specify defaults for variables

puremourning:
We want to be able to use `${Variable:default value goes here}`, but the built-in python template strings won't let us.

So looking at the standard library code, most of it is cruft we don't need. We can just pick and choose the code we want and re-implement, adding the syntax elements as we need them.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/215)

PR #216, opened 2020-07-23
Allow a statically configured list of gadgets

puremourning:
Useful for storing config in source control


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/216)

PR #218, opened 2020-07-24
Upgrade gadgets when they change

### LINKED ISSUE #176: [Feature Request]: Node 12+ support
Since today Node12 is the [current LTS](https://nodejs.org/en/about/releases/) (until the end of the year), I figure it would be nice to support those new versions.

Could you tell me the changes that limits the support to < 12? I'm ready to explore solutions but I don't understand the problem yet.

Thank you for your help

### LINKED ISSUE #217: [Feature Request]: Support NodeJS 12
[vscode-node-debug2](https://github.com/microsoft/vscode-node-debug2) has supported node>=8.0. 
Can Vimspector also support the same versions?

puremourning:
This adds a --upgrade option to install_gadget.py and makes
VimspectorUpdate only update things which have changed.

To do this, we record the gadget spec in a manfiest file and compare it
with the current spec when in upgrade mode.

'Changed' in this case means that the gadget spec has changed from the
last time the installer was run. It does _not_ actually check the
presence of the gadget.

--

Also update the node debugger to the latest version and remove the node 10 check :) 

Fixes #176 
Fixes #217

### DIFF
--- a/install_gadget.py
+++ b/install_gadget.py
@@ -69,6 +69,10 @@ parser.add_argument( '--force-all',
                      action = 'store_true',
                      help = 'Enable all unsupported completers' )
 
+parser.add_argument( '--upgrade',
+                     action = 'store_true',
+                     help = 'Only update adapters changed from the manifest' )
+
 parser.add_argument( '--quiet',
                      action = 'store_true',
                      help = 'Suppress installation output' )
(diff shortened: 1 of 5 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/218)

PR #219, opened 2020-07-24
Fixes issue with needed elevated privileges on Windows

lilwayne1556:
I ran into an issue with the installation on Windows where the mklink command was failing due to needing sufficient privileges and was able to fixed it by using a direction junction instead of a directory symbolic link

Installing netcoredbg...
You do not have sufficient privilege to perform this operation.
 - FAILED installing netcoredbg: Command '['cmd.exe', '/c', 'mklink', '/D', 'C:\\Users\\wayneb\\vimfiles\\plugged\\vimspector\\gadgets\\windows\\netcoredbg', 'C:\\Users\\wayneb\\vimfiles\\plugged\\vimspector\\gadgets\\windows\\download\\netcoredbg\\master\\root\\netcoredbg']' returned non-zero exit status 1

PR #220, opened 2020-07-24
Use a junction, as symlinks require escalated priviledges on some win…

puremourning:
…dows

closes #219


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/220)

PR #242, opened 2020-09-01
Allow customisation of the signs

### LINKED ISSUE #231: [Feature Request]: cutomizable sign priority
The issue is that Coc signs, Git signs and Vimspector signs are competing.
It happens that my Git signs priority is higher than Vimspector BP signs.

I'd like the signs priorities to be configurable like so:
```vim
let g:vimspectorBP_sign_priority = 10
let g:vimspectorBPDisabled_sign_priority = 0 
let g:vimspectorPC_sign_priority = 12
```

Modifying all the other sign priorities could work if the other plugins offer that option.
However, this is not really scallable.

puremourning:
Too many plugins use the default priority of 10 so they race/chase.
Allow uses to configure the priorities and make sure that the defaults
are documented.

Fixes #231

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,16 +66,18 @@ For a tutorial and usage overview, take a look at the
     * [Other servers](#other-servers)
  * [Customisation](#customisation)
     * [Changing the default signs](#changing-the-default-signs)
+    * [Sign priority](#sign-priority)
     * [Changing the default window sizes](#changing-the-default-window-sizes)
     * [Changing the terminal size](#changing-the-terminal-size)
     * [Advanced UI customisation](#advanced-ui-customisation)
+    * [Customising the WinBar](#customising-the-winbar)
     * [Example](#example)
  * [FAQ](#faq)
  * [Motivation](#motivation)
  * [License](#license)
  * [Sponsorship](#sponsorship)
 
-<!-- Added by: ben, at: Wed 22 Jul 2020 22:10:50 BST -->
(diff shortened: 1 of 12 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/242)

puremourning (commented):
:lgtm:

> Reviewed 12 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/242#-:-MG8z_uX7rPiS9iQqmM7:bnfp4nl)* status: all files reviewed, 1 unresolved discussion

___
*[python3/vimspector/breakpoints.py, line 386 at r1](https://reviewable.io/reviews/puremourning/vimspector/242#-MG8z6wl46sOjF7-c9VX:-MG8z6wl46sOjF7-c9VY:b1fkd9b) ([raw file](https://github.com/puremourning/vimspector/blob/c1b544fb3c4667a36a91661945aefb16e5a95a42/python3/vimspector/breakpoints.py#L386)):*
> ```Python
>                          'VimspectorBP',
>                          sign,
>                          9,
> ```

this is s alight change in behaviour, previously the priority here was 10 (while debugging).

It's tempting to put this back, but the override mechanism wouldn't allow to differentiate "while debugging" from not, so that makes it more tricky to describe. 

I'm going to leave this like it is because it makes describing the problem and the customisation simpler. if the original intention (i.e. higher priority when debugging) is still useful, we can add it back in.

PR #243, opened 2020-09-01
Don't leak buffers when creating output view

puremourning:
We create a new window for the output view, then we use _CreateBuffer to create some buffers. This leaks the empty buffer that gets put in the output view.

This solution re-uses that empty buffer for the first output buffer created.

TODO:

* [x] Check that VimspectorToggleLog doesn't leak one too
* [x] Check that the installer doesn't leak one too.

### DIFF
--- a/python3/vimspector/output.py
+++ b/python3/vimspector/output.py
@@ -17,6 +17,7 @@ from vimspector import utils, install
 
 import vim
 import json
+import typing
 
 
 class TabBuffer( object ):
@@ -55,6 +56,8 @@ def ShowOutputInWindow( win_id, category ):
 class OutputView( object ):
   """Container for a 'tabbed' window of buffers that can be used to display
   files or the output of commands."""
+  _buffers: typing.Dict[ str, TabBuffer ]
+
   def __init__( self, window, api_prefix ):
     self._window = window
     self._buffers = {}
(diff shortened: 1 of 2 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/243)

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/243)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[python3/vimspector/output.py, line 158 at r1](https://reviewable.io/reviews/puremourning/vimspector/243#-MG95TV95RP_CNgwwLkh:-MG95TV95RP_CNgwwLki:bja34a2) ([raw file](https://github.com/puremourning/vimspector/blob/d19cd57855608dd526af147c7902ec9588e4c768/python3/vimspector/output.py#L158)):*
> ```Python
>       # delete it. We could try and use it, but that complicates the call to
>       # SetUpCommandBuffer
>       buf_to_delete = self._window.buffer
> ```

I wonder if this could be any-old buffer in the event that we called `Clear()`. Though I suspect that would actually end up closing the window, because it deletes all the buffers that could be displayed in this window, unless this happens to be the current window.

Hmm. not sure how to detect this situation, perhaps self._window.buffer.name is None ?

puremourning (commented):
:lgtm:

> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/243#-:-MG96Bf74dtY019NJz7H:bnfp4nl)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @puremourning)

puremourning (commented):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/243)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #248, opened 2020-09-03
Support completion for console and watches.

### LINKED ISSUE #245: [Feature Request]: Code completion for Watch and Console windows
It would be nice to have code completion on "console" and "Watches" panes.

puremourning:
Add omnifunc for prompt buffers

This synchronous completion can be used with any completion system
including built-in CTRL-X CTRL-O.

The filetype of the prompt buffers is set to VimspectorPrompt so that it
can be identified by completion systems. For example, this works well
with YCM:

let g:ycm_semantic_triggers =  {
  \   'VimspectorPrompt': [ '.', '->', ':', '<' ]
  \ }

closes #245 

TODO:

* [x] docs and examples
* [ ] lots of tests
* [x] fix command line completion as this is behaving erratically
* [x] test with many other servers to validate assumptions about offsets
* [ ] add an asynchronous version for use with _other_ completion systems ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,9 +39,11 @@ For a tutorial and usage overview, take a look at the
     * [Stepping](#stepping)
     * [Variables and scopes](#variables-and-scopes)
     * [Watches](#watches)
+       * [Watch autocompletion](#watch-autocompletion)
     * [Stack Traces](#stack-traces)
     * [Program Output](#program-output)
        * [Console](#console)
+       * [Console autocompletion](#console-autocompletion)
        * [Log View](#log-view)
     * [Closing debugger](#closing-debugger)
  * [Debug adapter configuration](#debug-adapter-configuration)
@@ -77,7 +79,7 @@ For a tutorial and usage overview, take a look at the
  * [License](#license)
  * [Sponsorship](#sponsorship)
 
-<!-- Added by: ben, at: Tue  1 Sep 2020 13:42:32 BST -->
+<!-- Added by: ben, at: Fri  4 Sep 2020 00:48:17 BST -->
 
 <!--te-->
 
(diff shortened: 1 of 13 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/248)

puremourning:
it seems that the chrome debugger just returns the list and does not return 'start', but it does internally calculate one, so it doesn't work per the spec and therefore we probably need to mimic what vscode does (whatever that might be).

At least one other implementation I saw using something like `\k+$` to find the previous non-keyword character to use as `col`. maybe we should do that, but we're quickly stepping into completion-engine territory, which is very annoying territory for a debugger.

puremourning:
I'm pretty happy with this now. I think I'm going to merge it without tests because they will take ages to write and this is a really useful feature off the bat. There's plenty of other stuff that doesn't have tests yet. I'll try and add some over the next few days.

I'm [somewhat of an old hat](https://vimways.org/2019/a-test-to-attest-to/)  at testing insert-mode completion.

Though this case should be much easier as my implementation here is synchronous.

puremourning (commented):
:lgtm:

> Reviewed 3 of 5 files at r1, 10 of 10 files at r2.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/248#-:-MGLWz8iAgDjT2prlbV4:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

puremourning (commented):
:lgtm:

> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/248#-:-MGLYzUr6ABbOrgkI5GF:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

sloppydevil:
I'm using vimspector to debug chrome on typescript.
as to console and watch completion with coc-nvim, is there solutions already?

I set the filetype mandatorily to typescript and the completion won't do. But when I did that in other new file types without any LSP, coc-nvim will do.

PR #258, opened 2020-09-14
Alert user when vimspector.json is broken

Mte90:
I had some issues with the plugin until with a patch I discovered that my json was broken, this patch help to alter the user of those cases.

Probably the text can be written better and more clear.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/258)

puremourning:
Thanks for doing this. I have incorporated your change in another change and will merge that in due course.

For now, closing this PR.

PR #307, opened 2020-11-26
[WIP] Gadget refactor

puremourning:
Pull the gadget configs out into a "pluggable" API (sort of)..

This makes vimspector a python [namespace package](https://www.python.org/dev/peps/pep-0420/) and in theory [should allow us to just stick things in sys.path](https://packaging.python.org/guides/creating-and-discovering-plugins/#using-namespace-packages) but in practice, this doesn't work for vim's `python3/vimspector/plugins` directory because of vim's embedded python using a custom finder which [doesn't implement](https://github.com/vim/vim/blob/master/runtime/doc/if_pyth.txt#L367-L377) the required [iter_modules interface](https://docs.python.org/3.6/library/pkgutil.html#pkgutil.iter_modules)

in theory this could work if plugins were installed via pip install --user, though again vim sometimes has sys.path issues in that regard.

one option would be to forcefully scan through the `runtimepath` looking for vimspector directories or something, but I'm not sure what would work. for now the refactor is a start. let's see where it goes.

Looks like it might be possible to add an implementation for the built-in vim module to the iter_modules thing by [registering an implementaiton](https://github.com/python/cpython/blob/master/Lib/pkgutil.py#L143-L185)

### DIFF
--- a/install_gadget.py
+++ b/install_gadget.py
@@ -113,7 +113,7 @@ parser.add_argument( '--sudo',
                             "run this as root via sudo, pass this flag." )
 
 done_languages = set()
-for name, gadget in gadgets.GADGETS.items():
+for name, gadget in gadgets.Gadgets().items():
   lang = gadget[ 'language' ]
   if lang in done_languages:
     continue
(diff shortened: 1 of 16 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/307)

puremourning:
closing for now

Issue #310, opened 2020-12-01
Duplicate environment variables are send to vscode-cpptools on Windows OS.


zoumi:
Python will upcase all environment variables, so we can't copy all environment variables from os.environ and send they to vscode-cpptools.

Here is the link:
https://stackoverflow.com/questions/19023238/why-python-uppercases-all-environment-variables-in-windows

This is from vimspector source code(@debug_session.py line 742)
```python
    # TODO: Do we actually need to copy and update or does Vim do that?
    env = os.environ.copy()
    if 'env' in self._adapter:
      env.update( self._adapter[ 'env' ] )
    self._adapter[ 'env' ] = env
```

zoumi:
For example if we have an env var:
ProgramW6432=C:\Program Files

vimspector will send PROGRAMW6432 to vscode-cpptools.
But vscode-cpptools already has an env var call ProgramW6432 from it's father process. So it throws an Error.

zoumi:
Sorry for not using an issue template.This bug is obvious.

puremourning:
I can't confirm either of our proposed changes resolves this as you have not provided steps to reproduce. I have a windows vm and I have vimspector working in it, but I can't reproduce the problem. Please list steps to reproduce using the examples in vimspector/tests/testdata/cpp

zoumi:
We have to use the external console to reproduce it.
```
        "externalConsole": true,
```

Here is my .vimspector.json:
```json
{
  "configurations": {
    "Launch": {
      "adapter": "vscode-cpptools",
      "configuration": {
        "name":"my gdb debug",
        "type":"cppdbg",
        "request": "launch",
        "args": [],
        "env":[],
        "environment": [],
        "MIMode":"gdb",
        "program": "test.exe",
        "externalConsole": true,
        "miDebuggerPath": "gdb.exe"
      }
    }
  }
}
```

puremourning:
I can repro thanks. confirmed that latest master fixes this.

PR #311, opened 2020-12-01
Fix duplicate environment variables

zoumi:
Fix https://github.com/puremourning/vimspector/issues/310

### DIFF
--- a/python3/vimspector/debug_session.py
+++ b/python3/vimspector/debug_session.py
@@ -742,6 +742,8 @@ class DebugSession( object ):
 
     # TODO: Do we actually need to copy and update or does Vim do that?
     env = os.environ.copy()
+    if os.name == "nt":
+        env = {}
     if 'env' in self._adapter:
       env.update( self._adapter[ 'env' ] )
     self._adapter[ 'env' ] = env
--- a/python3/vimspector/debug_session.py
+++ b/python3/vimspector/debug_session.py
@@ -743,7 +743,7 @@ class DebugSession( object ):
     # TODO: Do we actually need to copy and update or does Vim do that?
     env = os.environ.copy()
     if os.name == "nt":
-        env = {}
+      env = {}
     if 'env' in self._adapter:
       env.update( self._adapter[ 'env' ] )
     self._adapter[ 'env' ] = env


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/311)

puremourning:
Thanks. Do you think we should just never copy the whole environment? Seems a bit wasteful anyway. IMO this is still a cpptools bug - the environment that we send is still _valid_. Though a few users have hit this now so thanks for the workaround.

zoumi:
Sorry, I'm not familiar with linux environment variables, so I test the OS name in my Code and test the code on my PC.
I think we don't need copy the environment variables  if you think it's OK to do so in linux.

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/311)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @zoumi)

___
*[python3/vimspector/debug_session.py, line 746 at r1](https://reviewable.io/reviews/puremourning/vimspector/311#-MNZhL4L2_cryCZH4O7R:-MNZhL4L2_cryCZH4O7S:b-dsotya) ([raw file](https://github.com/puremourning/vimspector/blob/b111e45bf5a5f94cd5a27a4a68f143d9909fdd79/python3/vimspector/debug_session.py#L746)):*
> ```Python
>     env = os.environ.copy()
>     if os.name == "nt":
>         env = {}
> ```

please fix lint error

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/311)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @zoumi)

___
*[python3/vimspector/debug_session.py, line 744 at r1](https://reviewable.io/reviews/puremourning/vimspector/311#-MNZh_YjEQRE5KHCSpM-:-MNZh_YjEQRE5KHCSpM0:b-xw8riv) ([raw file](https://github.com/puremourning/vimspector/blob/b111e45bf5a5f94cd5a27a4a68f143d9909fdd79/python3/vimspector/debug_session.py#L744)):*
> ```Python
> 
>     # TODO: Do we actually need to copy and update or does Vim do that?
>     env = os.environ.copy()
> ```

we should skip this copy if we're not going to use it

zoumi:
@puremourning Sorry, I can't sign in reviewable.io. 
May I close this request then you make a new request instead?

puremourning:
Thanks for this, I've pushed a similar patch #315 - can you try it?

PR #315, opened 2020-12-05
Don't copy the environment to update it

### LINKED ISSUE #310: Duplicate environment variables are send to vscode-cpptools on Windows OS.

puremourning:
This causes problems on windows, and is wasteful anyway. The subprocess
will pick up the environment from its parent.

Fixes #310
Closes #311 

@zoumi can you try this patch ? I took your idea and extended it for all platforms. This seems to work on macOS at least.

### DIFF
--- a/python3/vimspector/debug_session.py
+++ b/python3/vimspector/debug_session.py
@@ -740,11 +740,7 @@ class DebugSession( object ):
 
     self._connection_type = self._api_prefix + self._connection_type
 
-    # TODO: Do we actually need to copy and update or does Vim do that?
-    env = os.environ.copy()
-    if 'env' in self._adapter:
-      env.update( self._adapter[ 'env' ] )
-    self._adapter[ 'env' ] = env
+    self._adapter[ 'env' ] = self._adapter.get( 'env', {} )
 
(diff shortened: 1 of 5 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/315)

puremourning (commented):
:lgtm:

> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/315#-:-MNpAzyD4HGEytbO0Rhj:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #319, opened 2020-12-14
Fix docker example

puremourning:
The example was was using 'launchCommand' which is not valid according
    to the schema; it should be 'runCommand'.

    But also, it never really worked. Vimspector would start the "adapter"
    (in this case, try and connect to the TCP port) before running the
    "prepare" commands, wich in this case would actually start debugpy
    listening. So to solve that we run the prepare commands earlier.
    Hopefully this won't cause a regression for Java and C++ remote attach,
    which we don't really have tests for.

    Finally, due to the way docker works, when you forward a port and
    nothing is listening on it, docker _accepts_ the connection then
    immediately drops it. This is _super_ annoying meaning that it looks to
    vimspector liek the server instantly dies if it takes nonzero time for
    the remote commands to open the port. So to solve this we add loaunch
    and attach delays which can be configured in the adapter config. This
    actually solves a prolem where the java debugger just takes agest to
    attach on remote launch too.

    (Finally, finally...) updated the vimspector schema to represent
    the real launch/attach remote configuration, which was incorrectly
    spec'd at the adapter level, but it's actually per launch/attach block.

cc @evdoks

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -633,19 +633,24 @@ Vimspector then orchestrates the various tools to set you up.
           // Command to launch the debugee and attach the debugger; 
           // %CMD% replaced with the remote-cmdLine configured in the launch
           // configuration. (mandatory)
-          "launchCommmand": [
+          "runCommand": [
             "python", "-m", "debugpy", "--listen", "0.0.0.0:${port}",
             "%CMD%"
           ]
           
-          // Optional alternative to launchCommmand (if you need to run multiple
+          // Optional alternative to runCommand (if you need to run multiple
           // commands)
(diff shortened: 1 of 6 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/319)

puremourning (commented):
:lgtm:

> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/319#-:-MPAApep1vFa6H8Z2aXW:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #334, opened 2021-01-08
Fix node thread

### LINKED ISSUE #333: Debugging a node app has incorrect thread state
### Describe the bug

When starting debugging on a node js script, vimspector shows the main thread as running, but it is actually paused on entry. User must hit continue to start execution.

### Minimal reproduciton

> Please answer the following questions

* Does your issue reproduce using `vim -Nu /path/to/vimspector/support/minimal_vimrc` ? YES
* If you are using Neovim, does your issue reproduce using Vim? YES

> List of steps to reproduce:

1 vim -Nu minimal_vimrc support/test/node/simple/simple.js
2. F5

### Expected behaviour

Debugging starts and pauses on first line of script.

### Actual behaviour

Debugger shows no stack trace or PC, thread state is running.

From user @mkastner on Gitter.

### Additional info

The problem is the sequence of events sent by the debug adapter. Vimspector requests threads:

1. When the initialisation exchange completes (requests all threads)
2. Whenever a thread event is received 
3. whenever a  stopped event is received.

If any of those happens while any other request is in progress, we cache the request and handle it later. The latest request is processed when the response to the outstanding request is received. 

The problem is if the event is a stopped event, it is the handling of the threads request that actually sets the thread state internally to stopped. In a sequence where the first event is a stopped event, we end up discarding the stopped event. like:

1. Stopped event (thread 1 = stopped) (request threads)
2. Initialisation complete (cache request)
3. threads response received (discard response and process cached request)
4. response received (but forgotten about the stopped event).

The solution is to always process the thread response, but then handle the cached request. To avoid flicker, we don't draw the screen in the case where we're sending a cached request.

puremourning:
Fixes #333 

  Fix thread state and PC for starting node app

    The problem is the sequence of events sent by the debug adapter.
    Vimspector requests threads:

    * When the initialisation exchange completes (requests all threads)
    * Whenever a thread event is received
    * whenever a stopped event is received.

    If any of those happens while any other request is in progress, we cache
    the request and handle it later. The latest request is processed when
    the response to the outstanding request is received.

    The problem is if the event is a stopped event, it is the handling of
    the threads request that actually sets the thread state internally to
    stopped. In a sequence where the first event is a stopped event, we end
    up discarding the stopped event. like:

    1. Stopped event (thread 1 = stopped) (request threads)
    2. Initialisation complete (cache request)
    3. threads response received (discard response and process cached request)
    4. response received (but forgotten about the stopped event).

    The solution is to always process the thread response, even if we send
    the cached request. To avoid flicker, we don't draw the screen, or
    expand any threads/stacks in the case where we're sending a cached
    request.

### DIFF
--- a/python3/vimspector/stack_trace.py
+++ b/python3/vimspector/stack_trace.py
@@ -188,11 +188,12 @@ class StackTraceView( object ):
       return
 
     def consume_threads( message ):
+      requesting = False
       if self._requesting_threads == StackTraceView.ThreadRequestState.PENDING:
         # We may have hit a thread event, so try again.
         self._requesting_threads = StackTraceView.ThreadRequestState.NO
         self.LoadThreads( *self._pending_thread_request )
-        return
+        requesting = True
 
       self._requesting_threads = StackTraceView.ThreadRequestState.NO
       self._pending_thread_request = None
@@ -210,8 +211,6 @@ class StackTraceView( object ):
       if stopEvent is not None:
         stoppedThreadId = stopEvent.get( 'threadId' )
         allThreadsStopped = stopEvent.get( 'allThreadsStopped', False )
-
-      requesting = False
 
       # FIXME: This is horribly inefficient
       for t in message[ 'body' ][ 'threads' ]:
@@ -240,7 +239,10 @@ class StackTraceView( object ):
         # If this is a stopped event, load the stack trace for the "current"
         # thread. Don't do this on other thrads requests because some servers
         # just break when that happens.
-        if infer_current_frame:
+        #
+        # Don't do this if we're also satisfying a cached request already (we'll
+        # do it then)
+        if infer_current_frame and not requesting:
           if thread.id == self._current_thread:
             if thread.CanExpand():
               self._LoadStackTrace( thread, True, reason )
--- a/support/test/tcl/test
+++ b/support/test/tcl/test
@@ -3,6 +3,8 @@
 set SCALAR g
 array set ARRAY {key1 value1 key2 value2}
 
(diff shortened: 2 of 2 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/334)

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/334)* status: 0 of 2 files reviewed, 1 unresolved discussion

___
*[python3/vimspector/stack_trace.py, line 215 at r1](https://reviewable.io/reviews/puremourning/vimspector/334#-MQWlbERBv2HWQ0pFhy7:-MQWlbERBv2HWQ0pFhy8:bsbn34m) ([raw file](https://github.com/puremourning/vimspector/blob/36329ff16ad17141d8027787ec2f919f08e6c8fa/python3/vimspector/stack_trace.py#L215)):*
> ```Python
>         allThreadsStopped = stopEvent.get( 'allThreadsStopped', False )
> 
>       requesting = False
> ```

bug: this overwrites the previous value, so we do some pointless requests

puremourning (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/334#-:-MQWlxRI16DtzTfrljaO:bnfp4nl)* status: 0 of 2 files reviewed, 1 unresolved discussion

puremourning (commented):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/334)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #343, opened 2021-02-05
Send `User` event after enter/before leave buffer

przepompownia:


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/343)

przepompownia:
This user event helps me, for example, define local mapping on buffers visited on each step. I'm not sure is it a good name. 

At the moment I need only such event. Maybe an equivalent fired on buffer leaving will be also needed in the future (e.g. to restore some states disabled by autocommand using this event).

puremourning (commented):
Thanks for sending a PR, but we will need tests, documentation and example usage to go with it.

It's not obvious to me exactly how you would use this autocommand robustly. I would probably prefer something higher level, and/or just implementing the feature that you need rather than a user auto command (which many users won't understand). As you say, you probably need an event (at least) on, say 'VimspectorReset` to go with this, i.e. to do the unmap.

A number of users have requested something to map keys when starting debugging, and to unmap them when done. This isn't trivial, but I think I would prefer to implement that _full_ feature than just a small part of it.

> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/343#-:-MSoevmu3igowEgGHyqi:b-av01t7)* status: 0 of 1 files reviewed, all discussions resolved

przepompownia:
Thanks for the quick reply.

Initially I had local mappings defined only for `VimspectorUICreated` but these worked only for the buffer associated with the first step.

```vim
" vimspector-specific config file:
autocmd User VimspectorUICreated windo doautocmd User IDEDebuggerMapsNeeded
" config with abstract mappings:
autocmd User IDEDebuggerMapsNeeded call s:defineDebuggerMaps()
```
 (https://github.com/przepompownia/vim-arctgx/blob/master/bundleConfig/vimspector.vim#L22, https://github.com/przepompownia/vim-arctgx/blob/f793ee07647495bf93a6cbbe8c773143bf25f8c7/plugin/ide-map.vim#L4 - the project is nothing more than personal vimrc after refactor to keep maintaining easier).

Another reason is the need for mute linter messages in any form at the time of debugging.

I found how you test another autocommands based on events defined there and I can add something similar. 

I didn't want to do documentation without your approval of the idea itself.

przepompownia:
Describing the context I did not include the command I tested this change on:
```vim
autocmd User VimspectorFrameWasSet doautocmd User IDEDebuggerMapsNeeded
```
At the moment I added simple test (without jumping to another file). I have problems with executing `./run_tests` in local environment.

przepompownia:
I have some proposition of sending `VimspectorFrameLeavePre` event. The current version bases on a buffer variable `vimspectorStep` created after enter buffer (by `OpenFileInCurrentWindow()`) to check if the buffer was visited by Vimspector. 

Probably we have to check if the same buffer is still visited by another thread.

Because lack of time I did not write any test so far but I tested it manually:
```vim
autocmd User VimspectorFrameLeavePre echom printf('Leave %s!', bufname())
autocmd User VimspectorFrameEnter echom printf('Enter %s!', bufname())
```

puremourning (commented):
Sorry I'm even more confused now.

I think the desired functionality is that you are able to execute some code (such as to set mappings) in the context of a buffer whenever vimspector jumps the cursor to that buffer, and then run some similar code (such as to remove those mappings) when vimspector jumps the cursor to a different buffer?

I think if that's the case, we could simplify this to just a single event, raised after dong the jump: `VimspectorJumpedToFrame`, along with events raised when debugging starts `VimspectorDebugStarted` and when debugging ends `VimspectorDebugEnded`. The user's code can then record which buffers it has done mappings for and to add/remove mappings as necessary.

All in all, I think that's more general and more likely to be maintainable. You can imagine user code like

```viml

" Always use F5 for continue
nmap <F5> <Plug>VimspectorContinue

let s:mapped = {}

func! s:MapGeneral()
  " add global mappings, active while debugging
  nmap <F4> <Plug>VimspectorReset
  " ...
endfunc

function! s:MapForCurrentBuffer()
  let buffer = bufnr()
  if has_key( s:mapped, buffer )
    return
  endif

  " add the buffer-local mappings
  nmap <buffer> <F9> <Plug>VimspectorStepOver
  " ...

  let s:mapped[ buffer ] = 1
endfunc

func! s:Clear()
  " remove general mappings 
  nunmap <F4>

  for buf in keys( s:mapped )
    " remove the mappings
    nunmap <buffer> <F9>
    " ...
  endfor
endfunc

augroup VimspectorCustom
  au!
  au User VimspectorDebugStarted call s:MapGeneral()
  au User VimspectorJumpedToFrame call s:MapForCurrentBuffer()
  au User VimspectorDebugEnded call s:Clear()
augroup END

```

I think that while this is quite sophisticated, and many users simply won't understand it, the alternative of supporting this sort of thing "natively" would require quite a lot of vimspector code and possibly even a whole meta-language to describe which mappings to enable when/where. This would be a disaster, so I think the auto command approach is probably fine.

> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/343#-:-MSt7guV1nnMpfKTdQj2:b-63edd2)* status: 0 of 2 files reviewed, all discussions resolved

przepompownia:
Native support of saving/restoring any states specific to another plugins does not seem to be responsibility of Vimspector especially that these states could vary between saving and restoring it. 

Your approach is close to the second scenario that I wanted to propose as an alternative to the current. It also seems to be OK.

przepompownia:
I did not found how to execute any vim command on selected buffer without switching to it (like `setbufvar()` can with variables). Switching through several buffers would be annoying (and require to remember which buffer was the current before cleaning up).

puremourning:
Yeah good point. Hmmm. It’s tricky. In the python vimspector code we have ways of doing this temporary jumping to buffer etc. But it’s surprisingly tricky. However it is possible. 

I’m still torn on this one. I like the idea of this being a built in feature. But I dislike having a meta-mapping syntax. I also dislike having lots of complex autocommands. It’s really tricky to find an elegant solution that’s not a maintenance nightmare. 

🤔

przepompownia:
The only problem on the plugin side is sending events in the most practical way from the user's point of view. It seems to be very simple as the current proposition or one from those mentioned there. How complex the autocommands handling these events will be depends especially on the place the event was sent from.

przepompownia:
Offtopic: when you write about meta-mapping syntax do you mean `<Plug>(something)`? "Abstract mapping" seems to be good name as these are useful as an interface between physical shortcut and real action without need to force the user to apply the direct equivalent of such mapping.

puremourning:
> when you write about meta-mapping

Sorry I should have been more clear. When I say "Vimspector should support this natively", I'm thinking that as a user I should be able to say in my config "When starting debugging, map `{ these keys }` to `{ these functions }` then unmap them when I'm done debugging". This is what you're implementing with the custom user auto commands.

I think that the use case is common enough that my preference is that it's easy for the user to do. However, it's tricky to determine a simple configuration which defines the "map `{ these keys }` to `{ these functions }`" without essentially inventing a new language which amounts to a series of `:map` calls, which well, could just be an auto command, which is what you've proposed.

I'll look at this again when I have some time, but it may be that we end up just with the 2 user autocommands anyway :)

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/343)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @przepompownia)

___
*[python3/vimspector/utils.py, line 76 at r4](https://reviewable.io/reviews/puremourning/vimspector/343#-MTgX2vl4-kyiKyOvrF1:-MTgX2vl4-kyiKyOvrF2:bw3wgth) ([raw file](https://github.com/puremourning/vimspector/blob/219914ee8be08b2396d96eaf63b14fbd956201ac/python3/vimspector/utils.py#L76)):*
> ```Python
> 
> 
> def OpenFileInCurrentWindow( file_name ):
> ```

None of this code really belongs in here. It's making all sorts of assumptions about _why_ it is being called. A legit user could call `OpenFileInCurrentWindow` for any other purpose, not realising t has this very specific behaviour baked into it. We need to separate these concerns, and isolate the code relating to running the frame-jump auto commands to a context where that makes sense.
___
*[python3/vimspector/utils.py, line 79 at r4](https://reviewable.io/reviews/puremourning/vimspector/343#-MTgWVo67STgujxHc4Yb:-MTgWVo67STgujxHc4Yc:b499qd9) ([raw file](https://github.com/puremourning/vimspector/blob/219914ee8be08b2396d96eaf63b14fbd956201ac/python3/vimspector/utils.py#L79)):*
> ```Python
>   buffer_number = BufferNumberForFile( file_name )
>   try:
>     if "vimspectorStep" in vim.current.buffer.vars:
> ```

the problem with this is that there is no guarantee that the current buffer is the previous buffer that we called VimspectorFrameEnter for.

We should at least instead record which buffer that is (i.e. the stack frame _before_ changing) and temporarily switch to that buffer before running the user autocommand.

there's a proc `utils.LetCurrentBuffer()` and a few related procs used in a number of places to do a similar job.
___
*[python3/vimspector/utils.py, line 82 at r4](https://reviewable.io/reviews/puremourning/vimspector/343#-MTgXLJSCLU6HuYPzGvr:-MTgXLJSCLU6HuYPzGvs:b-dmospd) ([raw file](https://github.com/puremourning/vimspector/blob/219914ee8be08b2396d96eaf63b14fbd956201ac/python3/vimspector/utils.py#L82)):*
> ```Python
>       del vim.current.buffer.vars[ "vimspectorStep" ]
>       # @todo: what if the same buffer is still visited by another thread
>       vim.command( 'doautocmd <nomodeline> User VimspectorFrameLeavePre' )
> ```

it looks like we trigger the LeavePre and Enter auto commands on every frame jump. That means that we run a lot of redundant code when stepping though. We need to avoid this as that's the perf critical stuff (stepping)
___
*[tests/ui.test.vim, line 422 at r4](https://reviewable.io/reviews/puremourning/vimspector/343#-MTgXYYp8XqFTsgQPF1_:-MTgXYYp8XqFTsgQPF1a:b-m518vo) ([raw file](https://github.com/puremourning/vimspector/blob/219914ee8be08b2396d96eaf63b14fbd956201ac/tests/ui.test.vim#L422)):*
> ```vim
>   call assert_equal( 'foo', getbufvar(bufnr(), 'vimspectorStepIsThere', 0) )
> 
>   au! TestVimspectorFrameEnter
> ```

This hardly tests anything other than vim's built-in `au!` command. We need to test it being called _by vimspector_

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/343)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @przepompownia)

___
*[tests/ui.test.vim, line 422 at r4](https://reviewable.io/reviews/puremourning/vimspector/343#-MTgXYYp8XqFTsgQPF1_:-MTgXioNApF8jA_wwhZH:b-bcscgd) ([raw file](https://github.com/puremourning/vimspector/blob/219914ee8be08b2396d96eaf63b14fbd956201ac/tests/ui.test.vim#L422)):*

This hardly tests anything other than vim's built-in `au!` command. We need to test it being called _by vimspector_
</blockquote></details>

wait, sorry, I'm talking rubbish, ignore this.

puremourning (changes requested):
> Reviewed 1 of 3 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/343)* status: all files reviewed, 4 unresolved discussions (waiting on @przepompownia)

___
*[tests/ui.test.vim, line 420 at r4](https://reviewable.io/reviews/puremourning/vimspector/343#-MTgXqyh8cwCkYXS2Q2Q:-MTgXqyi2N_jF0cZNrAC:b-si72zp) ([raw file](https://github.com/puremourning/vimspector/blob/219914ee8be08b2396d96eaf63b14fbd956201ac/tests/ui.test.vim#L420)):*
> ```vim
> 
>   call s:StartDebugging()
>   call assert_equal( 'foo', getbufvar(bufnr(), 'vimspectorStepIsThere', 0) )
> ```

we need to step into other files and see that the au gets called with the right current buffer on the Enter and LeavePre events too.

przepompownia (commented):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/343)* status: all files reviewed, 4 unresolved discussions (waiting on @przepompownia and @puremourning)

___
*[python3/vimspector/utils.py, line 76 at r4](https://reviewable.io/reviews/puremourning/vimspector/343#-MTgX2vl4-kyiKyOvrF1:-MTgd8BVChkCFn9gA4Lg:bbzv604) ([raw file](https://github.com/puremourning/vimspector/blob/219914ee8be08b2396d96eaf63b14fbd956201ac/python3/vimspector/utils.py#L76)):*

None of this code really belongs in here. It's making all sorts of assumptions about _why_ it is being called. A legit user could call `OpenFileInCurrentWindow` for any other purpose, not realising t has this very specific behaviour baked into it. We need to separate these concerns, and isolate the code relating to running the frame-jump auto commands to a context where that makes sense.
</blockquote></details>

I guess that `SetCurrentFrame` is more correct place. 
___
*[python3/vimspector/utils.py, line 79 at r4](https://reviewable.io/reviews/puremourning/vimspector/343#-MTgWVo67STgujxHc4Yb:-MTgdOmN7po4WviXmPU5:b-hz88fa) ([raw file](https://github.com/puremourning/vimspector/blob/219914ee8be08b2396d96eaf63b14fbd956201ac/python3/vimspector/utils.py#L79)):*

the problem with this is that there is no guarantee that the current buffer is the previous buffer that we called VimspectorFrameEnter for.

We should at least instead record which buffer that is (i.e. the stack frame _before_ changing) and temporarily switch to that buffer before running the user autocommand.

there's a proc `utils.LetCurrentBuffer()` and a few related procs used in a number of places to do a similar job.
</blockquote></details>

Please explain more why it is problem. If the previous buffer wasn't associated with `VimspectorFrameEnter`, we have nothing to clean before leaving it. 
___
*[python3/vimspector/utils.py, line 82 at r4](https://reviewable.io/reviews/puremourning/vimspector/343#-MTgXLJSCLU6HuYPzGvr:-MTggGe-AV2KMtu6NcWz:b-qlv7pn) ([raw file](https://github.com/puremourning/vimspector/blob/219914ee8be08b2396d96eaf63b14fbd956201ac/python3/vimspector/utils.py#L82)):*

it looks like we trigger the LeavePre and Enter auto commands on every frame jump. That means that we run a lot of redundant code when stepping though. We need to avoid this as that's the perf critical stuff (stepping)
</blockquote></details>

If I understand, you'd like to avoid trigger leave/enter event multiple time on the same buffer (say, e.g, 50 steps between file1 and file2 ;-) ). It seems that the only solution is that proposed by you (with more hard cleaning on`VimspectorDebugEnded` instead of  `VimspectorFrameLeavePre`). Performance is a strong argument (even if heavy autocommands triggered by the new events comes from the user choice).

przepompownia:
I have still no idea how to write working test (i.e. how to simulate SetCurrentFrame).

The most of visited buffers is unlisted so we have less to clean at the end.

puremourning (changes requested):
I like the simplicity of the current approach! I'm going to try and write some user code to see if I can make it do what I think we want, then we can use that as an example in the docs (which we'll need to update)

> I have still no idea how to write working test (i.e. how to simulate SetCurrentFrame).

You can simulate the user jumping in different frames from within the stack trace window. something like:

```viml
" jump to the stack trace window
call win_gotoid( g:vimspector_session_windows.stack_trace )
# put the cursor on the stack frame you want to jump to
call cursor( .... )
# jump to it
call vimspector#GoToFrame()
```

You can also use breakpoints, which essentially does the same thing. there's  a multi-file example of this in `temporary_breakpoints.test.vim`

> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/343#-:-MU0GD_c4nvaa_PSXdOf:b-mkbsc2)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @przepompownia and @puremourning)

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/343)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @przepompownia and @puremourning)

___
*[python3/vimspector/code.py, line 137 at r7](https://reviewable.io/reviews/puremourning/vimspector/343#-MU0LlPGDgi_DAUWKGCY:-MU0LlPGDgi_DAUWKGCZ:b2wkool) ([raw file](https://github.com/puremourning/vimspector/blob/c1ff1a5adc4ac673379aabf0a8321b22dcb186a5/python3/vimspector/code.py#L137)):*
> ```Python
>     try:
>       utils.OpenFileInCurrentWindow( frame[ 'source' ][ 'path' ] )
>       vim.command( 'doautocmd <nomodeline> User VimspectorJumpedToFrame' )
> ```

We should do this after setting the cursor. That way the user can use the current cursor position in the auto command (if they need it)

puremourning (changes requested):
This quick example worked for me (using `support/custom_ui_vimrc`).. Nice.

```diff
diff --git a/support/custom_ui_vimrc b/support/custom_ui_vimrc
index 95c3360..4e9f635 100644
--- a/support/custom_ui_vimrc
+++ b/support/custom_ui_vimrc
@@ -1,5 +1,7 @@
 execute 'source' expand( '<sfile>:p:h' ) . '/minimal_vimrc'
 set noequalalways
+let mapleader = ','
+let maplocalleader = "\<Space>"

 function! s:CustomiseUI()
   let wins = g:vimspector_session_windows
@@ -59,11 +61,52 @@ function! s:CustomiseWinBar()
     nnoremenu WinBar.✕\ ᶠ⁸ :call vimspector#Reset()<CR>
 endfunction

+let s:mapped = {}
+
+function! s:OnJumpToFrame() abort
+  if has_key( s:mapped, string( bufnr() ) )
+    return
+  endif
+
+  nmap <silent> <buffer> <LocalLeader>dn <Plug>VimspectorStepOver
+  nmap <silent> <buffer> <LocalLeader>ds <Plug>VimspectorStepInto
+  nmap <silent> <buffer> <LocalLeader>df <Plug>VimspectorStepOut
+  nmap <silent> <buffer> <LocalLeader>dc <Plug>VimspectorContinue
+
+  let s:mapped[ string( bufnr() ) ] = 1
+endfunction
+
+function! s:OnDebugEnd() abort
+
+  let original_buf = bufnr()
+  let hidden = &hidden
+
+  try
+    set hidden
+    for bufnr in keys( s:mapped )
+      try
+        execute 'noautocmd buffer' bufnr
+        silent! nunmap <buffer> <LocalLeader>dn
+        silent! nunmap <buffer> <LocalLeader>ds
+        silent! nunmap <buffer> <LocalLeader>df
+        silent! nunmap <buffer> <LocalLeader>dc
+      endtry
+    endfor
+  finally
+    execute 'noautocmd buffer' original_buf
+    let &hidden = hidden
+  endtry
+
+  let s:mapped = {}
+endfunction
+
 augroup TestUICustomistaion
   autocmd!
   autocmd User VimspectorUICreated call s:CustomiseUI()
   autocmd User VimspectorTerminalOpened call s:SetUpTerminal()
   autocmd User VimspectorUICreated call s:CustomiseWinBar()
+  autocmd User VimspectorJumpedToFrame call s:OnJumpToFrame()
+  autocmd User VimspectorDebugEnded call s:OnDebugEnd()
 augroup END

 let g:vimspector_sign_priority = {
```

> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/343#-:-MU0OQGx6BejZyU6ArIY:b-zgbxsi)* status: 0 of 3 files reviewed, 4 unresolved discussions (waiting on @przepompownia and @puremourning)

puremourning (changes requested):
should probably use `mapset`/`maparg` to save any existing mapping and restore it too.

> Reviewed 4 of 4 files at r7.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/343#-:-MU0PPs08IV5XpWe0R2Y:bh5kj6a)* status: all files reviewed, 4 unresolved discussions (waiting on @przepompownia and @puremourning)

PR #344, opened 2021-02-06
Fix variable substitution for multiple defaulted vars

### LINKED ISSUE #340: Multiple defaulted-substitutions in one string does not work
Repro:

```
{
   "configurations": { "test": {
       "test": "${this:this} is a ${test:test}"
   } }
}
```

Expect `this` and `test` are vars with defaults `this` and `test` respectively.

Actual: invalid dollar error. It interprets the default value for `this` as `this} is a ${test:test` meaning that the regex is bugged.

There's a sort-of workaround but it's hacky. We can't even say "parse these variables in this order" so it's pot luck. One option as a quick workaround is to allow `variables` to be a list of objects. This would allow the expansions to happen first, then plug in the string. But really, the parser should be fixed (though I don't know how yet)

Also: add pytest tests for this stuff.

puremourning:
The problem was that the python regex engine is strictly left-to-right,
so matching `[^}]|\\}` against \\}} meant that the `\\` was consumed by
the left of the `|`. The solution is to just switch them around.

Also add a way to run python tests from within vim, so we can actually
test this stuff.

Fixes #340

### DIFF
--- a/python3/vimspector/utils.py
+++ b/python3/vimspector/utils.py
@@ -460,8 +460,8 @@ VAR_MATCH = re.compile(
       {(?P<braceddefault>               # or An {id:default} - default param, as
         (?P<defname>[_a-z][_a-z0-9]*)   #   an ID
         :                               #   then a colon
-        (?P<default>(?:[^}]|\})*)       #   then anything up to }, or a \}
-      )}                             |  #
+        (?P<default>(?:\\}|[^}])+)      #   then anything up to }, or a \}
+      )}                             |  #   then a }
       (?P<invalid>)                     # or Something else - invalid
     )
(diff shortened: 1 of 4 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/344)

puremourning (commented):
:lgtm:

> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/344#-:-MStEL8k2o0e12Q9cl3M:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #352, opened 2021-02-22
Fix restart when using CodeLLDB

puremourning:
Fix restarts always getting stuck "Initializing" when using CodeLLDB.

When using the restart command we re-use the configuration dict as-is,
so always re-use the same TCP port for the lldb socket. Originally it
was thought this was due to a race condition, having the port still
open, but it's not.

When doing a restart, or reset, we shut down the server after we get the
response to the disconnect message. CodeLLDB then also sends a
'terminated' message.

Previously we were forcefully closing the socket before killing the app,
after we get the 'disconnect' response. This meant that the OS buffer
for the socket to localhost:<the port> still contained the terminated
message at the point that we force-closed the socket and killed the
server.

The result was that the firt messages read from the "new" socket to
locahost:<port> were the last messagse written by the previous process,
trikcing vimspector into thinking that the server terminated the process
(before responding to the initialize request).

ANyway, the solution is to ensure that we read all messages from the
previous instance before considering it done. This is done by killing
the server if there is one *first* and then trying to read any messages
from the socket until it closes (reads EOF).

The tricky part is for when we didn't start the server (i.e. in a
multi-session setup). Here we simply _have_ to close the socket because
we can't know when we've received all of the messages, and we shouldn't
expect to receive any 'terminated' events after 'disconnect'.

### DIFF
--- a/autoload/vimspector/internal/channel.vim
+++ b/autoload/vimspector/internal/channel.vim
@@ -95,21 +95,44 @@ EOF
 endfunction
 
 function! vimspector#internal#channel#StopDebugSession() abort
-  if exists( 's:ch' ) && ch_status( s:ch ) ==# 'open'
+
+  if exists( 's:job' )
+    " We started the job, so we need to kill it and wait to read all the data
+    " from the socket
+
+    if job_status( s:job ) ==# 'run'
+      call job_stop( s:job, 'term' )
+    endif
+
+    while job_status( s:job ) ==# 'run'
+      call job_stop( s:job, 'kill' )
+    endwhile
+
+    unlet s:job
+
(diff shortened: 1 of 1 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/352)

puremourning (commented):
Unfortunately I have no idea how to fix this in the same way for neovim, which doesn't appear to have a ch_read() blocking equivalent, so I guess it has to remain broken.

:LGTM:

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/352#-:-MU8hVlY2kMIXhA9azo7:b748jmx)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #358, opened 2021-02-25
Allow zero or one args for :VimspectorEval (like :VimspectorWatch)

jez:
I looked for a test to cargo cult, but I couldn't find any place where
`:VimspectorWatch` was called with no args and fed a string to the
`input()` call.

I also looked at the commit that added the same behavior for
VimspectorWatch (4586aa36) and it included no tests. If there's an easy
way to test this, please let me know.

### DIFF
--- a/autoload/vimspector.vim
+++ b/autoload/vimspector.vim
@@ -249,23 +249,12 @@ function! vimspector#AddWatchPrompt( expr ) abort
   call vimspector#AddWatch( a:expr )
 endfunction
 
-function! vimspector#Evaluate( ... ) abort
+function! vimspector#Evaluate( expr ) abort
   if !s:Enabled()
     return
   endif
-
-  if a:0 == 0
-    let l:expr = input( 'Enter eval expression: ' , '', 'custom,vimspector#CompleteExpr' )
-  else
-    let l:expr = a:1
-  endif
-
-  if l:expr ==# ''
-    return
-  endif
-
   py3 _vimspector_session.ShowOutput( 'Console' )
-  py3 _vimspector_session.EvaluateConsole( vim.eval( 'l:expr' ), True )
+  py3 _vimspector_session.EvaluateConsole( vim.eval( 'a:expr' ), True )
 endfunction
 
 function! vimspector#EvaluateConsole( expr ) abort
--- a/autoload/vimspector.vim
+++ b/autoload/vimspector.vim
@@ -255,9 +255,13 @@ function! vimspector#Evaluate( ... ) abort
   endif
 
   if a:0 == 0
-    let l:expr = input( 'Enter eval expression: ' )
+    let l:expr = input( 'Enter eval expression: ' , '', 'custom,vimspector#CompleteExpr' )
   else
     let l:expr = a:1
+  endif
+
+  if l:expr ==# ''
+    return
   endif
 
   py3 _vimspector_session.ShowOutput( 'Console' )
--- a/autoload/vimspector.vim
+++ b/autoload/vimspector.vim
@@ -239,12 +239,19 @@ function! vimspector#AddWatchPrompt( expr ) abort
   call vimspector#AddWatch( a:expr )
 endfunction
 
-function! vimspector#Evaluate( expr ) abort
+function! vimspector#Evaluate( ... ) abort
   if !s:Enabled()
     return
   endif
+
+  if a:0 == 0
+    let l:expr = input( 'Enter eval expression: ' )
+  else
+    let l:expr = a:1
+  endif
+
   py3 _vimspector_session.ShowOutput( 'Console' )
-  py3 _vimspector_session.EvaluateConsole( vim.eval( 'a:expr' ), True )
+  py3 _vimspector_session.EvaluateConsole( vim.eval( 'l:expr' ), True )
 endfunction
 
 function! vimspector#EvaluateConsole( expr ) abort
--- a/plugin/vimspector.vim
+++ b/plugin/vimspector.vim
@@ -100,7 +100,7 @@ command! -bar -nargs=? -complete=custom,vimspector#CompleteOutput
 command! -bar
       \ VimspectorToggleLog
       \ call vimspector#ToggleLog()
-command! -bar -nargs=1 -complete=custom,vimspector#CompleteExpr
+command! -bar -nargs=? -complete=custom,vimspector#CompleteExpr
       \ VimspectorEval
       \ call vimspector#Evaluate( <f-args> )
 command! -bar
--- a/plugin/vimspector.vim
+++ b/plugin/vimspector.vim
@@ -100,7 +100,7 @@ command! -bar -nargs=? -complete=custom,vimspector#CompleteOutput
 command! -bar
       \ VimspectorToggleLog
       \ call vimspector#ToggleLog()
-command! -bar -nargs=? -complete=custom,vimspector#CompleteExpr
+command! -nargs=1 -complete=custom,vimspector#CompleteExpr
       \ VimspectorEval
       \ call vimspector#Evaluate( <f-args> )
 command! -bar
--- a/tests/variables.test.vim
+++ b/tests/variables.test.vim
@@ -508,6 +508,33 @@ function Test_EvaluateConsole()
   call vimspector#test#signs#AssertCursorIsAtLineInBuffer(
         \ 'vimspector.Console', len, v:null )
 
+  VimspectorEval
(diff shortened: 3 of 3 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/358)

puremourning (commented):
Please can you explain the use case?

> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/358#-:-MUN36Qm4boNkEgChhD1:b8wuji4)* status: 0 of 2 files reviewed, all discussions resolved

jez:
> Please can you explain the use case?

It looks like there's weird quoting or word splitting problems that happen when I try to evaluate my command. For example, in a C++ project using `vscode-cpptools`, when I type this:

```
:VimspectorEval -exec print printf("hello\n")
```

I see this output in the command window:

```
Evaluating: -exec print printf(
error: <user expression 0>:2:1: expected expression
;
^
error: <lldb wrapper suffix>:1:40: expected '}'
}                                      
                                       ^
<lldb wrapper prefix>:42:1: to match this '{'
{                                      
^
```

note that `-exec print printf(`. It looks like the start of the `"` character is causing something to get truncated (probably it's getting treated as a Vim comment character?)

I don't want to have to remember to escape every `"` character that I type when using `:VimspectorEval`. Using `input()` guarantees that the string I enter is passed verbatim to the debugger:

```
:VimspectorEval
Enter eval expression: -exec print printf("hello\n")
```

shows this output:

```
Evaluating: -exec print printf("hello\n")
(int) $0 = 6

@"hello\r\n"
```

puremourning:
Ok so the bug is probably escaping then. Thenn by problem with this approach is you lose all the completions and it’s sub optimal vs just using the prompt buffer. I think I’d rather fix the escalating issue. 

There are tests for the evaluation function here https://github.com/puremourning/vimspector/blob/master/tests/variables.test.vim#L473

jez:
> I think I’d rather fix the escaping issue

I'm not sure there's anything to fix? I think what's happening is that `:VimspectorEval` begins a valid Vim script line, and Vim script lines are allowed to have trailing comments, and those start with `"`.

> you lose all the completions

`input()` can be configured to take the same completion function as is used for the `:command` registration (`-complete=custom,vimspector#CompleteOutput`)

I can go ahead and get that wired up, so that it will not be a problem anymore.

> just using the prompt buffer

Does Neovim have prompt buffers? How does that work? Do I have to go into insert mode in the `Output` panel? Running `:VimspectorEval` was the only way I could evaluate an expression.

puremourning:
Neovim master has prompt buffers, I’m told. Entering insert mode, type stuff, hit CR... see help prompt-buffer and the vimspector README. 
 
However given I still have to support othervim 0.4, I still feel in my soul this should be fixable without downshifting to input(), but if not I guess I don’t have too much issue with it. I hadn’t realised that input could take the completion func. That’s interesting and worth trying out I guess.

jez:
> Neovim master has prompt buffers, I’m told.

Nice, I just downloaded the latest Neovim nightly build and confirmed that the prompt buffers showed up. Thanks for the tip.

> I still feel in my soul this should be fixable without downshifting to input(),

I spent some time reading the docs for `<f-args>` and `:command` to see if I was missing something, but I don't see anything that looks like an option we could pass to the `:command` invocation that changes this behavior.

If you have suggestions for something to try or to search for let me know.

> I hadn’t realised that input could take the completion func. That’s interesting and worth trying out I guess.

I got it working:

Screen Shot 2021-02-25 at 12 14 21 AM


> There are tests for the evaluation function here https://github.com/puremourning/vimspector/blob/master/tests/variables.test.vim#L473

Thanks for the pointer. I've added a test. 🤞 that it passes.

jez:
> Thanks for the pointer. I've added a test. 🤞 that it passes.

After a few tries, I got a passing test.

puremourning:
> After a few tries, I got a passing test.

You can run the tests locally: https://github.com/puremourning/vimspector/blob/master/CONTRIBUTING.md#pr-guidelines

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/358)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @jez)

___
*[autoload/vimspector.vim, line 258 at r4](https://reviewable.io/reviews/puremourning/vimspector/358#-MUNV_4i6qzkyx_LKE_u:-MUNV_4i6qzkyx_LKE_v:bedsuvd) ([raw file](https://github.com/puremourning/vimspector/blob/3726ded6dd86204ca56c2a8ab5980c143117d551/autoload/vimspector.vim#L258)):*
> ```
>    let l:expr = input( 'Enter eval expression: ' , '', 'custom,vimspector#CompleteExpr' )
> ```

please wrap 80 characters max.
___
*[tests/variables.test.vim, line 516 at r4](https://reviewable.io/reviews/puremourning/vimspector/358#-MUNX-xo9GaNqXdTv5sC:-MUNX-xo9GaNqXdTv5sD:b2e0o5h) ([raw file](https://github.com/puremourning/vimspector/blob/3726ded6dd86204ca56c2a8ab5980c143117d551/tests/variables.test.vim#L516)):*
> ```vim
> 
> 
> function Test_EvaluateInput()
> ```

Can we add a test which covers the original problem, i.e. use of double quotes and expressions with spaces?

I came to the tests to see if we can still do `:VimsepctorEval a + b + c` with `-nargs=?`

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/358)* status: 0 of 3 files reviewed, 3 unresolved discussions (waiting on @jez)

*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/358#-MUNXMlu-s2Lt5cyJQCd:-MUNXMlu-s2Lt5cyJQCe:byig2px) (no related file):*
Do we need any change to `README.md` to describe this case?

---

puremourning (commented):
> Reviewed 1 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/358)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @jez)

puremourning (commented):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/358)* status: all files reviewed, 3 unresolved discussions (waiting on @jez)

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/358)* status: all files reviewed, 4 unresolved discussions (waiting on @jez)

___
*[plugin/vimspector.vim, line 103 at r4](https://reviewable.io/reviews/puremourning/vimspector/358#-MUNXzU-6pP6LJHd6uWC:-MUNXzU-6pP6LJHd6uWD:bq9dxzs) ([raw file](https://github.com/puremourning/vimspector/blob/3726ded6dd86204ca56c2a8ab5980c143117d551/plugin/vimspector.vim#L103)):*
> ```vim
>       \ VimspectorToggleLog
>       \ call vimspector#ToggleLog()
> command! -bar -nargs=? -complete=custom,vimspector#CompleteExpr
> ```

If we remove `-bar` here, then I don't think we need the `input()`

jez:
> If we remove `-bar` here, then I don't think we need the `input()`

Nice find!

I've removed `-bar`, put everything else back to what it was, and changed the test so that it has both spaces and quote characters in it.

puremourning (approved):
nice!

:lgtm: thanks!

> Reviewed 3 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/358#-:-MUNaB8d4c97-I5XgbsR:bwbvgrn)* status: :shipit: complete! all files reviewed, all discussions resolved

jez:
I think the test as I've written it is platform specific, and will print different information on macOS (lldb) and Linux (gdb).

It's 2:00 a.m. here so I'm going to go to bed. If you want to push changes to fix the test, feel free. Otherwise I will get to it tomorrow or maybe the day after.

puremourning:
ok thanks. sleep well.

jez:
Ok I lied before I could fall asleep I had an idea for how I could make it platform agnostic.

Double check that the test still looks good to you but I think it's ready again.

puremourning:
@mergifyio squash

puremourning:
Manually merging to squash the commits

puremourning:
Thanks @jez !

PR #367, opened 2021-03-05
Refactor code quality issues

akshgpt7:
## Description
Hi, I ran [DeepSource](https://deepsource.io/) analysis on my fork of the repo, and found some interesting [code quality issues and bug risks](https://deepsource.io/gh/akshgpt7/vimspector/) that can be helpful here.
This PR fixes a few of the issues detected.

## Summary of changes
- Refactor unnecessary `else` / `elif` when `if` block has a `return` statement.
- Remove unnecessary lambda expressions.
- Remove usage of `assert` outside tests.
- Add `.deepsource.toml` file for continuous analysis on bug risks/performance/code-quality issues on new changes.

### DIFF
--- a/.deepsource.toml
+++ b/.deepsource.toml
@@ -1,0 +1,15 @@
+version = 1
+
+test_patterns = ["tests/**"]
+
+exclude_patterns = [
+    "**.vim",
+    "support/test/**"
+]
+
+[[analyzers]]
+name = "python"
+enabled = true
+
+  [analyzers.meta]
+  runtime_version = "3.x.x"
--- a/python3/vimspector/breakpoints.py
+++ b/python3/vimspector/breakpoints.py
@@ -292,7 +292,8 @@ class ProjectBreakpoints( object ):
 
 
   def SendBreakpoints( self, doneHandler = None ):
-    assert self._breakpoints_handler is not None
+    if self._breakpoints_handler is None:
+      raise AssertionError
 
     # Clear any existing breakpoints prior to sending new ones
     self._breakpoints_handler.ClearBreakpoints()
--- a/python3/vimspector/debug_adapter_connection.py
+++ b/python3/vimspector/debug_adapter_connection.py
@@ -218,7 +218,8 @@ class DebugAdapterConnection( object ):
 
     if len( self._buffer ) < content_length:
       # Need more data
-      assert self._state == 'READ_BODY'
+      if self._state != 'READ_BODY':
+        raise AssertionError
       return
 
     payload = str( self._buffer[ : content_length ], 'utf-8' )
--- a/python3/vimspector/debug_session.py
+++ b/python3/vimspector/debug_session.py
@@ -180,7 +180,7 @@ class DebugSession( object ):
               *shlex.split( response ),
               then = lambda: self.Start( new_launch_variables ) )
             return
-          elif response is None:
+          if response is None:
             return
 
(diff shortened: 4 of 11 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/367)

puremourning:
> Thanks for sending this. I took a look at the linked analysis and, well, I was disapointed.

There are, for example, some things which are truly useful (sucha as [this hilarious bug, which I thought I had fixed](https://deepsource.io/gh/akshgpt7/vimspector/issue/PYL-W0104/occurrences)), but then a _most_ of them are just opinionated style things which I don't actually agree with, for example:

* https://deepsource.io/gh/akshgpt7/vimspector/issue/PYL-R1720/occurrences 
* https://deepsource.io/gh/akshgpt7/vimspector/issue/PYL-R1711/occurrences
* https://deepsource.io/gh/akshgpt7/vimspector/issue/PYL-W0612/occurrences
* https://deepsource.io/gh/akshgpt7/vimspector/issue/PYL-W0613/occurrences

So in order for this tool to be genuinely useful, we'd have to have some way to suppress all the noise and stuff that's not representing any real issue, or stuff that represents backward or opinionated style things that are irrelevant. An example of which is:

> A single return or return None statement is found at the end of function or method definition. This statement can safely be removed to improve the readability because Python will implicitly return None.

The fact is that removing the statements does _not_ improve readability, it does the opposite, because it requires that the reader know the implicit behaviours. In @puremourning  style guide, functions which return a value, should conclude with a `return <value>` statement.


So looking at the ones that we've changed in this PR :

> Refactor unnecessary else / elif when if block has a return statement.

I don't see this as being significant improvement.

> Remove unnecessary lambda expressions.

This seems ok, though the implicit `self` in python I find a bit unusual.

> Remove usage of assert outside tests.

I don't agree with this one at all, the alternative is both more verbose and less expressive. I also didn't see it in the report - I was going to try and read the description of the issue to understand why this is being changed.

> Add .deepsource.toml file for continuous analysis on bug risks/performance/code-quality issues on new changes.

As mentioned, tis would need some tuning to remove a lot of noise. WIth a good set of reports, this could be added to the project as part of the required checks, but for now it's just too noisy. It's also not possible to run locally, so the "PR Failed, sigh I have to change something random" loop happens, which can be annoying.

akshgpt7:
Hey @puremourning, sorry for the late response.


> So in order for this tool to be genuinely useful, we'd have to have some way to suppress all the noise and stuff that's not representing any real issue, or stuff that represents backward or opinionated style things that are irrelevant

We have designed the tool by keeping this particular thing in mind.
The issues that you feel are opinionated, or conflict with `puremorning` styleguide, you can ignore them on different levels just with a click:
Screenshot (44)

Additionally, you can also configure what kind of issues you want to see and what kind issues should make your PRs red.
Screenshot (45)

> Remove usage of assert outside tests.

This issue is raised for `assert` statement in non-test files.
Python has an option to compile the optimized bytecode and create the respective `.pyo` files by using the options `-O` and `-OO`.
During this, all the assert statements are removed.
Now, it is never intended to run the optimized bytecode, this issue can be ignored as well. But, it is still recommended not to have it in non-test files.
A better way for internal self-checks is to explicitly check and raise respective `Error` using an `if` statement.

akshgpt7:
@puremourning any updates on this?

PR #368, opened 2021-03-08
Not able to set breakpoint in java libraries

cosminadrianpopescu:
For some lsp services, like java, the compiler can open the source files from various libraries. For java, if I have the source code, the code will be opened by go to definition in a virtual buffer, which does not exists locally (jdt://contents/<library-name>). 

Setting the break point then in library will fail silently. 

This PR fixes this issue. First, I did a check if the file exists physically, in which case, the key remains the same. However, if the file does not exists physically, then the key will be the exact file name.

### DIFF
--- a/python3/vimspector/breakpoints.py
+++ b/python3/vimspector/breakpoints.py
@@ -152,7 +152,9 @@ class ProjectBreakpoints( object ):
 
 
   def _PutLineBreakpoint( self, file_name, line, options ):
-    self._line_breakpoints[ os.path.abspath( file_name ) ].append( {
+    abs_name = os.path.abspath( file_name )
+    path = abs_name if os.path.isfile( abs_name ) else file_name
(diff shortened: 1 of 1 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/368)

puremourning:
We need to add a test, otherwise I'll likely break this in future, as this issue does not occur for me. 

https://github.com/puremourning/vimspector/blob/master/CONTRIBUTING.md#pr-guidelines

cosminadrianpopescu:
For a test, I think the java project needs to be modified to include an external library with sources. I will have a look at how that test goes, when I will have time.

puremourning:
I decided to include this anyway while I was thinking about something related, thanks for doing it and sorry for not merging it earlier. I'll include the patch with slight modification when merging "saved sessions" #465

PR #374, opened 2021-03-12
Make any key press dismiss the keyboard popup

puremourning:
Otherwise it kinda traps the keyboard annoyingly.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/374)

puremourning:
This doesn't work.

`popup_filter_menu()` always returns 1:

```
    // consume all keys until done
    rettv->v_type = VAR_BOOL;
    rettv->vval.v_number = VVAL_TRUE;
    res.v_type = VAR_NUMBER;
```

(Ok it returns 0 specifically when dragging with the mouse, but that's beside the point)

PR #437, opened 2021-08-02
breakpoint menu

dsych:
Hey @puremourning here is another tasty PR with long promised breakpoint view!

# Some highlights:
* Removed `quickfix` and implemented our own little window for more control
* Breakpoint view is **reactive**, meaning it will dynamically change based on modifications to any breakpoints
* Function breakpoints are now visible in this view!
* You can manage breakpoint right from inside the view: `c` for clear - remove breakpoint (multiple function breakpoints are removed based on function name); `t` for toggle - toggles breakpoint state (function breakpoints are simply removed).
* `<Enter>` or `<2-LeftMouse>` on any entry (except for function breakpoint) to jump to window where that file is loaded. If file is not opened, a new split will be opened with the correct file loaded.
* Regular navigation
* Right now relies on the old logic for generating quickfix list, but could be modified.

# Other things that might be useful:
* Implement menu with buttons for **toggle** and **clear** just like watch window?
* Behaviour when debugging is in process. Maybe have it as another split like watch view?
* Refactor code and breakpoint logic and move everything inside breakpoint class? Will definitely need your help. **EDIT**: DONE!
* **Tests** (LOL) will implement once we stabilize the api.

> NOTE: I am all in on `neovim` now, I can't test on vim. I tried to used only compatible apis, though.

## Demo
Peek 2021-08-04 15-25


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/437)

puremourning:
wow thanks so much for this! I'll try and set aside some time to look at it soon. sorry quite busy at the moment. Please do ping me if you don't get any updates by, say next week!

dsych:
A couple of other ideas we could pursue:
* display breakpoint symbol instead of words `ENABLED`, `DISABLED`
* adding new breakpoints right inside the view?
* adding highlight groups to differentiate parts of each line?

dsych:
ping @puremourning

puremourning (changes requested):
Thanks. Took this for a spin, first impressions - really nice. Took a first pass review on the code below too.

> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/437#-:-MglIXt22HEHFa-l-P_J:bmjb2fr)* status: 0 of 3 files reviewed, 8 unresolved discussions (waiting on @dsych)

*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/437#-MglIZSe0f7e3TbTeHtR:-MglIZSe0f7e3TbTeHtS:blmqfsd) (no related file):*
There's a bug when the window is already open, but in another tab. Steps to repro:

1. dtvim support/test/cpp/simple_c_program/test_c.cpp
2. set a breakpoint (or don't I guess)
3. call vimspector#ListBreakpoints()
4. start debugging (F5)
5. pick a profile
6. with the debug tab open, call vimspector#ListBreakpoints()

Expect:
 - breakpoints window opened on current tab

Actual:
 - current window is closed!

---
*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/437#-MglJWdeBTITTiXutwQT:-MglJWdeBTITTiXutwQU:b9wtxxo) (no related file):*
should we add the breakpoints window id to g:vimspector_session_windows, and maybe fire some user autocommand to enable customisation: https://github.com/puremourning/vimspector/#advanced-ui-customisation

---
___
*[python3/vimspector/breakpoints.py, line 62 at r3](https://reviewable.io/reviews/puremourning/vimspector/437#-MglKGGJ6Atr7BPWhinQ:-MglKGGJ6Atr7BPWhinR:b-ihoydz) ([raw file](https://github.com/puremourning/vimspector/blob/4fe9d29c4aeba2e80dd837965dd50247444434a9/python3/vimspector/breakpoints.py#L62)):*
> ```Python
>         self._breakpoint_win.buffer, "vimspector.Breakpoints"
>       )
>       mappings = [
> ```

we should probably reverse these and get then from settings, as in `mappings = settings.Dict( 'mappings' )[ 'breakpoints' ]` - this would work then like the others.

```

  # Mappings
  'mappings': {
    'variables': {
      'expand_collapse': [ '<CR>', '<2-LeftMouse>' ],
      'delete': [ '<Del>' ],
      'set_value': [ '<C-CR>', '<leader><CR>' ]
    },
    'stack_trace': {
      'expand_or_jump': [ '<CR>', '<2-LeftMouse>' ],
      'focus_thread': [ '<leader><CR>' ],
    }
  },

```
___
*[python3/vimspector/breakpoints.py, line 84 at r3](https://reviewable.io/reviews/puremourning/vimspector/437#-MglKkUR4QoD9VWePZya:-MglKkUR4QoD9VWePZyb:b-v9k4yf) ([raw file](https://github.com/puremourning/vimspector/blob/4fe9d29c4aeba2e80dd837965dd50247444434a9/python3/vimspector/breakpoints.py#L84)):*
> ```Python
>     self._breakpoint_list = breakpoint_list
> 
>     breakpoint_list = list(
> ```

nit maybe, but I would like to keep with my existing style which is to only break when necessary, and never bin pack arguments.

so

```python
breakpoints_list = list( map( _formatEntry, breakpoint_list ) )
```

in other places, I see 

```python
        utils.SetBufferContents(
          self._breakpoint_win.buffer, breakpoint_list
        )

```

but this should be like this if it fits on one line

```
        utils.SetBufferContents( self._breakpoint_win.buffer, breakpoint_list )
```

or this if it doesn't:

```
        utils.SetBufferContents( self._breakpoint_win.buffer,
                                 breakpoint_list )

```

or, if a single argument doesn't fit, then:

```
        utils.SetBufferContents(
          self._breakpoint_win.buffer,
          breakpoint_list )
```

it should be very rare (if ever) to have a `)` on its own line.
___
*[python3/vimspector/breakpoints.py, line 99 at r3](https://reviewable.io/reviews/puremourning/vimspector/437#-MglLl0wEbGQjpt3kGZU:-MglLl0wEbGQjpt3kGZV:beyye3z) ([raw file](https://github.com/puremourning/vimspector/blob/4fe9d29c4aeba2e80dd837965dd50247444434a9/python3/vimspector/breakpoints.py#L99)):*
> ```Python
>   def CloseBreakpoints( self ):
>     if self._breakpoint_win and self._breakpoint_win.valid:
>       vim.command( "{}close".format( self._breakpoint_win.number ) )
> ```

we need to store and use the tab page number (see `utils.LetCurrentTabPage`), or use the window ID for this. 
___
*[python3/vimspector/breakpoints.py, line 116 at r3](https://reviewable.io/reviews/puremourning/vimspector/437#-MglM43s4BE0NBx8Ezok:-MglM43s4BE0NBx8Ezol:b-lzelul) ([raw file](https://github.com/puremourning/vimspector/blob/4fe9d29c4aeba2e80dd837965dd50247444434a9/python3/vimspector/breakpoints.py#L116)):*
> ```Python
> 
>   def ToggleBreakpointView( self, breakpoint_list ):
>     if self._breakpoint_win and self._breakpoint_win.valid:
> ```

we should consider whether we want one of these per tab page or just one globally. It would be tempting to have an instance of this class per tab-page then only invoke methods on the one for the current tab page, however, this is super annoying because the vim auto commands for tab changing make it really hard to know when a tab is closed, etc. maybe we can work with that, or just change the semantics of "toggle" to mean: If the window is visible on another tab page, move the window here"
___
*[python3/vimspector/breakpoints.py, line 203 at r3](https://reviewable.io/reviews/puremourning/vimspector/437#-MglOBdo1fQ2vG6lJzE-:-MglOBdp5nv9z5g0U6-c:b-svczf) ([raw file](https://github.com/puremourning/vimspector/blob/4fe9d29c4aeba2e80dd837965dd50247444434a9/python3/vimspector/breakpoints.py#L203)):*
> ```Python
>     bp = self._breakpoints_view.GetBreakpointForLine()
>     if bp and bp.get( 'type' ) != 'F':
>       isSuccess = int(
> ```

we don't use camelCase variables here. a better name is `success`. but TBH, we can ditch the var and use `if int( utils.Call( 'win_gotoid', bp[ 'filename' ]' ) ):`

similar elsewhere, I know it's a little less efficient, but I kind of prefer using `utils.Call` rather than "naked" `vim.eval` for simple functions. It ends up more readable, with less string-faffing.
___
*[python3/vimspector/breakpoints.py, line 236 at r3](https://reviewable.io/reviews/puremourning/vimspector/437#-MglMcn5CMXBtJU76Lt8:-MglMcn5CMXBtJU76Lt9:b-4hnhar) ([raw file](https://github.com/puremourning/vimspector/blob/4fe9d29c4aeba2e80dd837965dd50247444434a9/python3/vimspector/breakpoints.py#L236)):*
> ```Python
>             json.dumps( bp[ 'options' ] ) )
>         } )
>     # I think this shows that the qf list is not right for this.
> ```

not sure this comment is useful anymore?

We can/should probably remove the quick fix list legacy from these functions.

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/437)* status: 0 of 3 files reviewed, 10 unresolved discussions (waiting on @dsych)

*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/437#-MglPFjs9f47MKBgKMoG:-MglPFjs9f47MKBgKMoH:bz16x59) (no related file):*
reminder: need to update the tests, obviously

---
*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/437#-MglPHM45ASYcyGaX5IF:-MglPHM45ASYcyGaX5IG:b5bx3v2) (no related file):*
reminder: need to update the README/docs

---

dsych (commented):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/437)* status: 0 of 3 files reviewed, 10 unresolved discussions (waiting on @dsych and @puremourning)

*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/437#-MglIZSe0f7e3TbTeHtR:-Mgvvieo6sxUJmO-WN3B:b1kxy83) (no related file):*

There's a bug when the window is already open, but in another tab. Steps to repro:

1. dtvim support/test/cpp/simple_c_program/test_c.cpp
2. set a breakpoint (or don't I guess)
3. call vimspector#ListBreakpoints()
4. start debugging (F5)
5. pick a profile
6. with the debug tab open, call vimspector#ListBreakpoints()

Expect:
 - breakpoints window opened on current tab

Actual:
 - current window is closed!
</blockquote></details>

good catch, fixed

---
___
*[python3/vimspector/breakpoints.py, line 62 at r3](https://reviewable.io/reviews/puremourning/vimspector/437#-MglKGGJ6Atr7BPWhinQ:-Mgvvcyo8kyNNSjG9KQu:b-n9tkhd) ([raw file](https://github.com/puremourning/vimspector/blob/4fe9d29c4aeba2e80dd837965dd50247444434a9/python3/vimspector/breakpoints.py#L62)):*

we should probably reverse these and get then from settings, as in `mappings = settings.Dict( 'mappings' )[ 'breakpoints' ]` - this would work then like the others.

```

  # Mappings
  'mappings': {
    'variables': {
      'expand_collapse': [ '<CR>', '<2-LeftMouse>' ],
      'delete': [ '<Del>' ],
      'set_value': [ '<C-CR>', '<leader><CR>' ]
    },
    'stack_trace': {
      'expand_or_jump': [ '<CR>', '<2-LeftMouse>' ],
      'focus_thread': [ '<leader><CR>' ],
    }
  },

```
</blockquote></details>

changed
___
*[python3/vimspector/breakpoints.py, line 99 at r3](https://reviewable.io/reviews/puremourning/vimspector/437#-MglLl0wEbGQjpt3kGZU:-MgvvWUXDkZwWC2YuERo:bjqarug) ([raw file](https://github.com/puremourning/vimspector/blob/4fe9d29c4aeba2e80dd837965dd50247444434a9/python3/vimspector/breakpoints.py#L99)):*
> utils.LetCurrentTabPage

is super nifty! used it
___
*[python3/vimspector/breakpoints.py, line 116 at r3](https://reviewable.io/reviews/puremourning/vimspector/437#-MglM43s4BE0NBx8Ezok:-MgvunMW2c6nu4pcbsAb:byjsv0e) ([raw file](https://github.com/puremourning/vimspector/blob/4fe9d29c4aeba2e80dd837965dd50247444434a9/python3/vimspector/breakpoints.py#L116)):*

we should consider whether we want one of these per tab page or just one globally. It would be tempting to have an instance of this class per tab-page then only invoke methods on the one for the current tab page, however, this is super annoying because the vim auto commands for tab changing make it really hard to know when a tab is closed, etc. maybe we can work with that, or just change the semantics of "toggle" to mean: If the window is visible on another tab page, move the window here"
</blockquote></details>

went with your suggestion of maintaining a single view across all tabs. it's a lot simpler from the code point of view. otherwise, the next best option IMO is using `win_execute`, which is not available in nvim (`nvim_win_call` is the alternative). hence, we would need to implement an abstraction and this is just too much work :D
___
*[python3/vimspector/breakpoints.py, line 203 at r3](https://reviewable.io/reviews/puremourning/vimspector/437#-MglOBdo1fQ2vG6lJzE-:-Mgvu4bnELhcx21Y_c-f:b-52j6zw) ([raw file](https://github.com/puremourning/vimspector/blob/4fe9d29c4aeba2e80dd837965dd50247444434a9/python3/vimspector/breakpoints.py#L203)):*

we don't use camelCase variables here. a better name is `success`. but TBH, we can ditch the var and use `if int( utils.Call( 'win_gotoid', bp[ 'filename' ]' ) ):`

similar elsewhere, I know it's a little less efficient, but I kind of prefer using `utils.Call` rather than "naked" `vim.eval` for simple functions. It ends up more readable, with less string-faffing.
</blockquote></details>

i will correct the camelCase variable name, but using `utils.Call` is a tricky one. how do i know if my call is complex or not? i can see it being used in very few place except for the utils themselves... i think it is best to leave it as is

dsych (commented):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/437)* status: 0 of 5 files reviewed, 10 unresolved discussions (waiting on @dsych and @puremourning)

*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/437#-MglJWdeBTITTiXutwQT:-MgxW0DM7Z1kJQWIgNKb:b-896fix) (no related file):*

should we add the breakpoints window id to g:vimspector_session_windows, and maybe fire some user autocommand to enable customisation: https://github.com/puremourning/vimspector/#advanced-ui-customisation
</blockquote></details>

Done.

---
___
*[python3/vimspector/breakpoints.py, line 84 at r3](https://reviewable.io/reviews/puremourning/vimspector/437#-MglKkUR4QoD9VWePZya:-Mgw-DzfBjCJeC-76sKQ:blu390f) ([raw file](https://github.com/puremourning/vimspector/blob/4fe9d29c4aeba2e80dd837965dd50247444434a9/python3/vimspector/breakpoints.py#L84)):*

nit maybe, but I would like to keep with my existing style which is to only break when necessary, and never bin pack arguments.

so

```python
breakpoints_list = list( map( _formatEntry, breakpoint_list ) )
```

in other places, I see 

```python
        utils.SetBufferContents(
          self._breakpoint_win.buffer, breakpoint_list
        )

```

but this should be like this if it fits on one line

```
        utils.SetBufferContents( self._breakpoint_win.buffer, breakpoint_list )
```

or this if it doesn't:

```
        utils.SetBufferContents( self._breakpoint_win.buffer,
                                 breakpoint_list )

```

or, if a single argument doesn't fit, then:

```
        utils.SetBufferContents(
          self._breakpoint_win.buffer,
          breakpoint_list )
```

it should be very rare (if ever) to have a `)` on its own line.
</blockquote></details>

hopefully fixed everywhere... have you thought about introducing autoformatter? this would save you a ton of time during reviews. i believe they can also integrate with your linting rules
___
*[python3/vimspector/breakpoints.py, line 203 at r3](https://reviewable.io/reviews/puremourning/vimspector/437#-MglOBdo1fQ2vG6lJzE-:-Mgvxvhu6LufDq-GEzcL:b-ibqamh) ([raw file](https://github.com/puremourning/vimspector/blob/4fe9d29c4aeba2e80dd837965dd50247444434a9/python3/vimspector/breakpoints.py#L203)):*
<details><summary><i>Previously, dsych (Dmytro Sych) wrote…</i></summary><blockquote>

i will correct the camelCase variable name, but using `utils.Call` is a tricky one. how do i know if my call is complex or not? i can see it being used in very few place except for the utils themselves... i think it is best to leave it as is
</blockquote></details>

also, this is not just a call to `win_gottoid` there is also a nested call to `bufwinid`, so in reality this would have looked more like:
```py
utils.Call("win_gotoid", utils.Call("winbufid", bp['filename']))
``` 
___
*[python3/vimspector/breakpoints.py, line 236 at r3](https://reviewable.io/reviews/puremourning/vimspector/437#-MglMcn5CMXBtJU76Lt8:-Mgw-mIjEnoLaclbtPl_:bp9ecrg) ([raw file](https://github.com/puremourning/vimspector/blob/4fe9d29c4aeba2e80dd837965dd50247444434a9/python3/vimspector/breakpoints.py#L236)):*

not sure this comment is useful anymore?

We can/should probably remove the quick fix list legacy from these functions.
</blockquote></details>

i actually found it quite useful when determining the location of breakpoint during jump. honestly, we could leave it as is

dsych:
@puremourning there is still a need to sync the breakpoint sign during debugging:
> say our current pointer is at line X and there is also a breakpoint there. we need to modify the point to reflect the existanse of the breakpoint. 

the problem is that breakpoint logic and pointer logic are now split between two modules and there is no easy way to let one or the other know about the type of sign it needs to display without breaking encapsulation.
i really need your opinion on this, since i might be modifying the underlining framework to accomodate async sign updates:
1. we could have a separate module that is responsible to sign rendering and all other modules will just delegate.
2. we could have a subscription based model that i described earlier, where different modules will subscribe/emit events when there is a need to update signs
3. ? you are welcome to suggest your option.

puremourning:
I don’t fully understand the problem but on the face of it it seems simple - just have the thing that renders the PC ask the breakpoints thing if there’s a breakpoint in the line. 
Avoid unnecessary abstraction at all costs.

dsych:
Right, however both PC and breakpoint could be updated independently. Consider the case when you toggle a breakpoint at PC

puremourning:
I see. We could still have the code window be the ‘sign controller’ during debugging but essentially delegate the breakpoints part to the ‘breakpoint’ module. When not debugging we could have some other thing be the current ‘sign controller’. We could have debug session marshal which one is in control.

dsych:
I am worried that would fracture the sign controller, which would lead us to the same situation we are in with code and breakpoint windows right now...

puremourning:
ok, so going back to your original suggestion. I think you suggested something like a "signs" module which the breakpoints and the code window both use to set signs. Something like

breakpoints.py

```
   self.signs.SetBreakpointAtLine( file, line, style ) # maybe return a sign ID ?
   self.signs.RemoveBreakpointAtLine( ... ) # maybe take a sign ID ?
```

code.py

```
    self.signs.SetPCAtLine( file, line ) 
    self.signs.MovePCToLine( file, line ) # shrug
```

Then `signs.py` would have logic like to determine if both breakpoint and PC are on the same line? We'd also have to think about the signs in the stack trace window, but that isn't so important. My worry with this is that it's yet-more-copies-of-the-state which is easy to mess up.

would that be materially better ? Hard for me to be sure without doing it and seeing what it feels like. One suggestion is to write the code in breakpoints.py and code.py in the most natural way, that will tell you the API you need for the new thing. It's common to do this backwards (i.e. design the API then try and use it; it's always better to write the usage code first, or at least draft it so you know what feels intuitive).

dsych:
so after much deliberation i've decided to try event based arch where breakpoint and code modules could subscribe to an event emitter( hand rolled, very simple) and update their respective signs. here is how it works:
* if there are any changes coming from breakpoints module, it will emit event and code module will update its sings.
* on the other hand, if we move a PC through code module, nothing really happens for the breakpoint module, since code's sings have higher precedence.   
* if breakpoint is toggled at PC's line, breakpoint module sets its own sing and notifies code module, which overwrites current line's sign with a higher precedence sing of its own.

After some manual testing, things seem to be working as expected, but of course test are failing *spectacularly*, since sign groups are no longer inside code module and ids do not align.
I would appreciate some help fixing the tests, since i dont know what most of them are doing)

dsych:
@puremourning i fixed all of the existing tests (and updated the once related to the old implementation of breakpoint view with location list) and synced with upstream master. Once you have a chance, i would appreciate a round of review.

oh one more thing, when we stabilize the api, i will update the readme and add more tests for things like mappings, cross-tab opening etc.

puremourning:
Thanks! I’ll try and look this weekend. I have a free day (ish) on Saturday \o/

puremourning (changes requested):
I'm generally happy with the API and approach, so we can look at polish now too.

Sorry for the review delay. 

> Reviewed 1 of 9 files at r8.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/437#-:-MlKq1lo3dUz5gn29gwT:bt1cz8s)* status: 1 of 11 files reviewed, 35 unresolved discussions (waiting on @dsych and @puremourning)

*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFqGTK5rFSvnIeIuh9:-MlFqGTK5rFSvnIeIuhA:b-v9f4hp) (no related file):*
I managed to cause a nasty bug I think with temporary breakpoints. Steps to repro:

* gh pr checkout 437
* edit support/test/cpp/simple_c_program/.vimspector.json and change `StopAtEntry` to `StopOnEntry` for CodeLLDB
* edit support/test/cpp/simple_c_program/test_c.cpp
* F5
* you should be presented with the assembly in pre-main (i.e. this has triggered a `source` request and we're presented with a temporary buffer containing that source)
* F9
* Traceback

Here's a vid: https://asciinema.org/a/HDBJMVbLE1igdwkJI9BVll6qF


Her's the traceback:

```
Error detected while processing function <SNR>108__OnServerData:
line    7:
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/Users/ben/.vim/bundle/vimspector/python3/vimspector/debug_session.py", line 376, in OnChannelData
    self._connection.OnData( data )
  File "/Users/ben/.vim/bundle/vimspector/python3/vimspector/debug_adapter_connection.py", line 158, in OnData
    self._ReadBody()
  File "/Users/ben/.vim/bundle/vimspector/python3/vimspector/debug_adapter_connection.py", line 241, in _ReadBody
    self._OnMessageReceived( message )
  File "/Users/ben/.vim/bundle/vimspector/python3/vimspector/debug_adapter_connection.py", line 264, in _OnMessageReceived
    request.handler( message )
  File "/Users/ben/.vim/bundle/vimspector/python3/vimspector/breakpoints.py", line 525, in <lambda>
    lambda msg, source=source, temp_idxs=temp_idxs: response_handler(
  File "/Users/ben/.vim/bundle/vimspector/python3/vimspector/breakpoints.py", line 481, in response_handler
    self._UpdateTemporaryBreakpoints( breakpoints, temp_idxs )
  File "/Users/ben/.vim/bundle/vimspector/python3/vimspector/breakpoints.py", line 423, in _UpdateTemporaryBreakpoints
    f"{ user_bp[ 'line' ] } to { bp[ 'line' ] }" )
KeyError: 'line'
```

---
___
*[python3/vimspector/breakpoints.py, line 84 at r3](https://reviewable.io/reviews/puremourning/vimspector/437#-MglKkUR4QoD9VWePZya:-MlFcC7y893zavRfSaaE:b-zd2t5b) ([raw file](https://github.com/puremourning/vimspector/blob/4fe9d29c4aeba2e80dd837965dd50247444434a9/python3/vimspector/breakpoints.py#L84)):*
<details><summary><i>Previously, dsych (Dmytro Sych) wrote…</i></summary><blockquote>

hopefully fixed everywhere... have you thought about introducing autoformatter? this would save you a ton of time during reviews. i believe they can also integrate with your linting rules
</blockquote></details>

thought about, somewhat. done anything about ? no. there are some custom flake8 checks that we use to check for these things but nothing that actually changes the code. I'm still not sure that's something I actually want though tbh. code is for reading mainly, and computers (even good ones) aren't good at making judgement calls.
___
*[python3/vimspector/breakpoints.py, line 116 at r3](https://reviewable.io/reviews/puremourning/vimspector/437#-MglM43s4BE0NBx8Ezok:-MlFjvd92zGOetk18LXk:buq1v0v) ([raw file](https://github.com/puremourning/vimspector/blob/4fe9d29c4aeba2e80dd837965dd50247444434a9/python3/vimspector/breakpoints.py#L116)):*
<details><summary><i>Previously, dsych (Dmytro Sych) wrote…</i></summary><blockquote>

went with your suggestion of maintaining a single view across all tabs. it's a lot simpler from the code point of view. otherwise, the next best option IMO is using `win_execute`, which is not available in nvim (`nvim_win_call` is the alternative). hence, we would need to implement an abstraction and this is just too much work :D
</blockquote></details>

one globally is great. tried it out and it feels intuitive.
___
*[python3/vimspector/breakpoints.py, line 203 at r3](https://reviewable.io/reviews/puremourning/vimspector/437#-MglOBdo1fQ2vG6lJzE-:-MlFmS23C65zb07KQPKb:b-fgdi0j) ([raw file](https://github.com/puremourning/vimspector/blob/4fe9d29c4aeba2e80dd837965dd50247444434a9/python3/vimspector/breakpoints.py#L203)):*
<details><summary><i>Previously, dsych (Dmytro Sych) wrote…</i></summary><blockquote>

also, this is not just a call to `win_gottoid` there is also a nested call to `bufwinid`, so in reality this would have looked more like:
```py
utils.Call("win_gotoid", utils.Call("winbufid", bp['filename']))
``` 
</blockquote></details>

ok this is fine
___
*[python3/vimspector/breakpoints.py, line 70 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFbiKZBAAiDR4S1G8g:-MlFbiKZBAAiDR4S1G8h:b-o8icpy) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/breakpoints.py#L70)):*
> ```Python
>                      'vimspector#JumpToBreakpointViewBreakpoint()<CR>' )
>       # set highlighting
>       vim.eval( "matchadd( 'WarningMsg', 'ENABLED', 100 )" )
> ```

we should ideally use the highlight group associated with the diagnostic sign. not sure that's particularly easy to do, though, so this is fine.
___
*[python3/vimspector/breakpoints.py, line 75 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFbzWVDsHSirlWrayA:-MlFbzWVDsHSirlWrayB:b-nozga4) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/breakpoints.py#L75)):*
> ```Python
> 
>       # we want to maintain the height of the window
>       self._breakpoint_win.options[ "winfixheight" ] = True
> ```

what case was this catching ? we don't do this for (e.g.) the output window I don't think.
___
*[python3/vimspector/breakpoints.py, line 99 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFdSr01VGh4qidC-Cf:-MlFdSr01VGh4qidC-Cg:bo9jzcx) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/breakpoints.py#L99)):*
> ```Python
>       return None
> 
>     line_num = int( vim.eval( 'getpos( \'.\' )' )[ 1 ] )
> ```

I think `vim.current.window.cursor[0]` would be better here
___
*[python3/vimspector/breakpoints.py, line 192 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFm5et56FiqHkTFu0g:-MlFm5et56FiqHkTFu0h:b-w1ti0u) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/breakpoints.py#L192)):*
> ```Python
>     bp = self._breakpoints_view.GetBreakpointForLine()
>     if bp:
>       if bp.get( 'type' ) == 'F':
> ```

can we create some symbolic names for these types?

e.g.

class BreakpointType:
   FUNCTION = 'F'
   LINE = 'L'

...

use BreakpointType.FUNCTION...
___
*[python3/vimspector/breakpoints.py, line 195 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFmL1I3GVUH2NiEjmV:-MlFmL1I3GVUH2NiEjmW:b-9m4bn0) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/breakpoints.py#L195)):*
> ```Python
>         self.ClearFunctionBreakpoint( bp.get( 'filename' ) )
>       else:
>         self._ToggleBreakpoint( None, bp.get( 'filename' ), bp.get( 'lnum' ),
> ```

newlines after `,`
___
*[python3/vimspector/breakpoints.py, line 205 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFmgb--x7AzYFoRMCa:-MlFmgb--x7AzYFoRMCb:baa0a6i) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/breakpoints.py#L205)):*
> ```Python
> 
>       if not success:
>         vim.command( "leftabove split {}".format( bp.get( 'filename' ) ) )
> ```

using `bp.get()` here makes it seem like `None` is a valid result. If it resurnds None, then I think you'll just get a nasty vim traceback (same below)

either, `filename` must be in there (and use `bp[ 'filename' ]` or guard against it. Same for `lnum`
___
*[python3/vimspector/breakpoints.py, line 207 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFmtZ64jrNozOPW-lr:-MlFmtZ64jrNozOPW-ls:b-6xhoz6) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/breakpoints.py#L207)):*
> ```Python
>         vim.command( "leftabove split {}".format( bp.get( 'filename' ) ) )
> 
>       vim.eval( "setpos( '.', [0, {}, 1, 1] )".format( bp.get( 'lnum' ) ) )
> ```

prefer to use `vim.current.window.cursor = ( bp[ 'lnum'], 0 )` (or better, create `utils.SetCurrentLineAndColumn( lnum, 1 )` see comment in code.py about line being 1 based and column being 0-based).
___
*[python3/vimspector/breakpoints.py, line 270 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFnxaX8vH7dXDwXTH3:-MlFnxaX8vH7dXDwXTH4:bhg8mwz) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/breakpoints.py#L270)):*
> ```Python
>     return None, None
> 
>   def _FindVerifiedBreakpoint( self, breakpoint_id ):
> ```

strictly speaking does this really find only verified breakpoints ? It seems to find only those with an ID, which IIRC might have not been verified. e.g. where the breakpoint is on an invalid line?
___
*[python3/vimspector/breakpoints.py, line 287 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFoBJBD7iAUCnyNuej:-MlFoBJCEbVAfBMHiP6A:bkew2es) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/breakpoints.py#L287)):*
> ```Python
> 
>     # we are just updating position of the existing breakpoint
>     bp[ 'line' ]  = breakpoint.get( 'line' )
> ```

is it legal for this to be None ? if not, then we have to either default value or use breakpoint[ 'line' ]

I like to only use `.get( x )` where `x` can legitimately be missing. I'd rather have a runtime crash at the point it's extracted if it's missing unexpectedly rather than some-time-later when the value is used from `bp['line']`
___
*[python3/vimspector/breakpoints.py, line 298 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFohHV-DLRV7s-qo4N:-MlFohHV-DLRV7s-qo4O:bm3o2ra) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/breakpoints.py#L298)):*
> ```Python
>         continue
> 
>       self._PutLineBreakpoint( source.get( 'path' ), breakpoint.get( 'line' ),
> ```

break after `,`, single hanging indent
___
*[python3/vimspector/breakpoints.py, line 322 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFrNRmAiNOMdnHTdVl:-MlFrNRmAiNOMdnHTdVm:brmxj38) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/breakpoints.py#L322)):*
> ```Python
>       'line': line,
>       'options': options,
>       'id': id,
> ```

when we save to the session file, we must make sure to pop this out of the dict (we don't want to store it)
___
*[python3/vimspector/breakpoints.py, line 323 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFrS193LNERooXWfZg:-MlFrS193LNERooXWfZh:b-rubrb7) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/breakpoints.py#L323)):*
> ```Python
>       'options': options,
>       'id': id,
>       'path': path
> ```

I'm less clear if we want to store this in the session file. Also, we need to consider reading from existing session files that don't contain this.
___
*[python3/vimspector/breakpoints.py, line 346 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFtsfN60YSkQlTWk_y:-MlFtsfN60YSkQlTWk_z:bckqo33) ([raw file](https://github.com/puremourning/vimspector/blob/1ff9acb47ff4fc0a7432ed5c6e8a64c4afc992fb/python3/vimspector/breakpoints.py#L346)):*
> ```Python
>         if 'sign_id' in bp:
>           signs.UnplaceSign( bp[ 'sign_id' ], 'VimspectorBP' )
>           del bp[ 'sign_id' ]
> ```

unclear why this was removed?
___
*[python3/vimspector/breakpoints.py, line 357 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFtps-BRGoxTNC7H5g:-MlFtps-BRGoxTNC7H5h:b9kiyb1) ([raw file](https://github.com/puremourning/vimspector/blob/1ff9acb47ff4fc0a7432ed5c6e8a64c4afc992fb/python3/vimspector/breakpoints.py#L357)):*
> ```Python
>         dap_bp.pop( 'temporary', None )
> 
>         if bp[ 'options' ].get( 'temporary' ):
> ```

was this removed by accident?
___
*[python3/vimspector/breakpoints.py, line 631 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFtyMkAY2QKUlMgJ1V:-MlFtyMkAY2QKUlMgJ1W:buae1ed) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/breakpoints.py#L631)):*
> ```Python
> 
>   def Refresh( self ):
>     # TODO: Just this file ?
> ```

we should remove this comment. I though I had removed it recently, put perhaps not.
___
*[python3/vimspector/breakpoints.py, line 655 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFuCDEEb9KMBwzomJk:-MlFuCDEEb9KMBwzomJl:b-fgwpyl) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/breakpoints.py#L655)):*
> ```Python
>   def Load( self, save_data ):
>     self.ClearBreakpoints()
>     self._line_breakpoints = save_data.get( 'line', {} )
> ```

do we need to patch up where the `'path'` key is missing ?
___
*[python3/vimspector/code.py, line 24 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFurS02KrAT3ymiMMy:-MlFurS02KrAT3ymiMMz:b-d0pmdp) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/code.py#L24)):*
> ```Python
> 
> class CodeView( object ):
>   def __init__( self, window, api_prefix, render_event_emitter,
> ```

newlines after `,`, align arguments
___
*[python3/vimspector/code.py, line 92 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFv2C8DvGYFHP-gmLL:-MlFv2C8DvGYFHP-gmLM:bkl4hq3) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/code.py#L92)):*
> ```Python
>       return False
> 
>     abs_path = os.path.abspath( file_path )
> ```

should this be Breakpoints.NormalisePath ? If so, we can make that a module-scope function and so it's usable from here.
___
*[python3/vimspector/debug_session.py, line 68 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFwMBB4Y6E75YTo4pu:-MlFwMBB4Y6E75YTo4pv:borogiq) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/debug_session.py#L68)):*
> ```Python
>     self._outputView = None
>     self._codeView = None
>     self._breakpoints = breakpoints.ProjectBreakpoints( self._render_emitter,
> ```

break after `(` single hanging indent
___
*[python3/vimspector/debug_session.py, line 799 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFwlIu-ZDGLQZZl98J:-MlFwlIu-ZDGLQZZl98K:b6bl0em) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/debug_session.py#L799)):*
> ```Python
>     code_window = vim.current.window
>     self._codeView = code.CodeView( code_window, self._api_prefix,
>       self._render_emitter, self._breakpoints.IsBreakpointPresentAt )
> ```

split at `,`, single hanging indent
___
*[python3/vimspector/debug_session.py, line 857 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFwzY01LTN2Cx1Pek9:-MlFwzY01LTN2Cx1PekA:bz0onul) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/debug_session.py#L857)):*
> ```Python
>     # Code window
>     code_window = vim.current.window
>     self._codeView = code.CodeView( code_window, self._api_prefix,
> ```

don't bin pack arguments
___
*[python3/vimspector/settings.py, line 69 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFxLuD-LQgRf_nVka3:-MlFxLuD-LQgRf_nVka4:bovs0b7) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/settings.py#L69)):*
> ```Python
>     'breakpoints': {
>       'toggle': [ 't', '<F9>' ],
>       'delete': [ 'c', 'd' ],
> ```

I feel like `<Del>` is a good choice here too.
___
*[python3/vimspector/utils.py, line 873 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlG-hKS4BQdSuCFb88F:-MlG-hKS4BQdSuCFb88G:b-h2gmms) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/python3/vimspector/utils.py#L873)):*
> ```Python
> 
> 
> class Subject( object ):
> ```

I found this kind of confusing, and maybe somewhat over engineered.

I think what this all does is:

- the debug session has a `EventEmitter`
- the breakpoints manager is passed a pointer to it
- so is the code window
- the breakpoints thing 'subscribes' to it and gets a `Subject` back (I found it hard to understand the purpose of this, but I think the class should just be called `Subscription`)
- so does the code window
- this 'subscribe' operation adds some callback to the 'emitter'
- the 'subscription' object is just a pointer to the emitter and an ID (which so far as I can see is not used)
- at some point the code window calls unsubscribe on the subscription, which removes the callback from the emitter
- at some point the code or breakpoints window call 'emit' which triggers the 2 callbacks in arbitrary order

It seems to me that most of this is unnecessary abstraction and all we really need is a list of things that currently care about rendering signs. I dunno, it feels like this would be simpler and more on the nose:

```python
class Event:
    def __init__( self ):
        self.callbacks = set()

    def Subscribe( self, callback ):
        self.callbacks.put( callback )
    
   def Unsubscribe( self, callback ):
        self.callbacks.remove( callback )

    def Fire( self, *args, **kwargs ):
        for c in callbacks:
            c( *args, **kwargs )

```

WDYY?
___
*[tests/breakpoints.test.vim, line 44 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFy9F_D78rx8Uu5OZ8:-MlFy9F_D78rx8Uu5OZ9:bfe2bb) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/tests/breakpoints.test.vim#L44)):*
> ```vim
>   call vimspector#ClearBreakpoints()
>   call vimspector#test#signs#AssertSignGroupEmpty( 'VimspectorBP' )
>   call vimspector#test#signs#AssertSignGroupEmpty( 'VimspectorBP' )
> ```

this is duplicating the above call now.
___
*[tests/breakpoints.test.vim, line 440 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFyGCy9dAeO7sq17ik:-MlFyGCy9dAeO7sq17il:b-jr3js0) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/tests/breakpoints.test.vim#L440)):*
> ```vim
>         \ v:null )
> 
>   " FIXME: Display of braekpoints while debugging is kinda broken. we should
> ```

I think you fixed this FIXME ?!
___
*[tests/breakpoints.test.vim, line 579 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFyktc7FDQg9lq8yZ-:-MlFyktc7FDQg9lq8yZ0:b7k3sax) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/tests/breakpoints.test.vim#L579)):*
> ```vim
> " endfunction
> 
> function! s:CheckQuickFixEntries( entry )
> ```

I'm kind of confused. this isn't really checking QuickFixEntries anymore. Can we name it after what it does and name the vars more to what they really are.

This feels more like it's really similar to something like this (used elsewhere in the tests):

```
  call WaitForAssert( {->
        \   AssertMatchList(
        \     [
          \       'i is 1',
          \       'i is 3',
          \       'i is 5',
          \       'i is 7',
        \     ],
        \     GetBufLine( winbufnr( g:vimspector_session_windows.output ), -3 )
        \   )
        \ } )
```
___
*[tests/breakpoints.test.vim, line 584 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFyLt7ECyCpxJYRUNk:-MlFyLt7ECyCpxJYRUNl:bfo8v65) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/tests/breakpoints.test.vim#L584)):*
> ```vim
>   if len( a:entry ) != len( lines )
>     call assert_report(string(a:entry).' '.string(lines))
>     call assert_report( 'Expected to have '.len( a:entry ).' breakpoints, but got '.len( lines ) )
> ```

looks dangerously like this is > 80 characters
___
*[tests/breakpoints.test.vim, line 590 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFyhzR35CkucWLw3FK:-MlFyhzR35CkucWLw3FL:bsvwx2f) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/tests/breakpoints.test.vim#L590)):*
> ```vim
>     for key in a:entry
>       if assert_equal(key,
>                        \ lines[ i ] ) == 1
> ```

don't think this line break is necessary
___
*[tests/breakpoints_doublewidth.test.vim, line 544 at r10](https://reviewable.io/reviews/puremourning/vimspector/437#-MlFzYZk-sQ2Na0AOcOY:-MlFzYZl1eH5Fjl_0gyd:bmca4c2) ([raw file](https://github.com/puremourning/vimspector/blob/783015cc7cc6fb80a6075fafcc7f9fea93a42c86/tests/breakpoints_doublewidth.test.vim#L544)):*
> ```vim
> " endfunction
> 
> function! s:CheckQuickFixEntries( entry )
> ```

same comments as the other file

puremourning (commented):
> Reviewed 1 of 3 files at r2, 4 of 9 files at r8, 5 of 5 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/437)* status: all files reviewed, 35 unresolved discussions (waiting on @dsych and @puremourning)

dsych:
@puremourning thanks for the review and apologies for the delayed response. I will have some time to take a look at it on the weekend.

puremourning:
Merged in #531

PR #455, opened 2021-09-12
Open debugger in same tab

sheerun:
It seems not necessary to start debugger in separate tab. A better approach to me seems to be just using the same tab

### DIFF
--- a/python3/vimspector/debug_session.py
+++ b/python3/vimspector/debug_session.py
@@ -418,7 +418,6 @@ class DebugSession( object ):
       self._variablesView.Reset()
       self._outputView.Reset()
       self._codeView.Reset()
-      vim.command( 'tabclose!' )
       vim.command( 'doautocmd <nomodeline> User VimspectorDebugEnded' )
       self._stackTraceView = None
       self._variablesView = None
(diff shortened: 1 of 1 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/455)

puremourning:
Thanks for the contribution, but on the face of of it, sorry I completely disagree. I feel like this change would create a broken UI in any but the most trivial of cases. If you were to simply make a split in the initial tab for example, the UI would be pretty messed up. I think this would also break the wide/narrow mode selection and I’m extremely confident that it would break a merge number of the tests. 

How extensively has this been tested?

sheerun:
Maybe introduce it as an option? I don't use tabs or splits in Vim so it works in my case. There's nothing to test as the layout is the same

sheerun:
Or maybe automatically detect if there are no splits or tabs present, and use this method

puremourning:
I would only consider an option if there were a significant number road users that would use this. Honestly I don’t imagine that it’s worth it. 

My motto is ‘if there’s an option to not have an option, take the no-option option.’

sheerun:
Ok I guess I'll fork then

puremourning:
> Or maybe automatically detect if there are no splits or tabs present, and use this method

Again I do t see the benefit of this. As it is a workflow, script, and customisation breaking change there would have to be an extremely strong reason for that.

PR #473, opened 2021-10-02
Adhoc Launc Configs

jpfeiffer16:
The thought here is to add the option to be able to launch with an ad-hoc config passed when you call the launch function.

The use case here is integrating debugging for unit tests in C# in the Omnisharp-Vim plugin.

I made a nominal effort to get the tests updated, but I was missing some go binaries and lua stuff, so hopefully those work in the CI runs. If not I will look at those.

Please let me know if you'd like me to do anything differently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -800,6 +800,30 @@ See [our YouCompleteMe integration guide](#usage-with-youcompleteme) for
 another example where it can be used to specify the port to connect the [java
 debugger](#java---partially-supported)
 
+To launch with an ad-hoc config you can use:
+
+* `call vimspector#LaunchWithConfigurations( dict )`
+
+The argument is a `dict` wich is the `configurations` section of a .vimspector file
+Pass one configuration in and that will be selected as the one to run.
+For example:
+
+```viml
+   let pid = <some_exression>
+   call vimspector#LaunchWithConfigurations({
+               \  "attach": {
+               \    "adapter": "netcoredbg",
+               \    "configuration": {
+               \      "request": "attach",
+               \      "processId": proc_id
+               \    }
+               \  }
+               \})
+```
+
+This would launch the debugger and attach to the specified process without the need
+to have a local .vimspector file on disk.
+
 ### Debug configuration selection
 
 Vimspector uses the following logic to choose a configuration to launch:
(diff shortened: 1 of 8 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/473)

puremourning:
Thanks! If you have docker you should be able to run the tests in the container  https://github.com/puremourning/vimspector/blob/master/CONTRIBUTING.md#running-the-tests-locally

jpfeiffer16:
Thanks, I will do that

aqez:
@jpfeiffer16 just out of curiosity, are you currently able to debug the entire suite of tests without this functionality? If so what does your config look like? I’ve been trying to get dotnet test to attach unsuccessfully

jpfeiffer16:
@aqez . This is more focused on debugging individual tests/tests in a class.

You should be able to do what you are looking for with a little bit of manual work as things stand right now.

create a config that looks like this:
```
{
  "configurations": {
    "attach": {
      "adapter": "netcoredbg",
      "configuration": {
        "request": "attach",
        "processId": "${pid}"
      }
    },
    ....
}

```

run `VSTEST_HOST_DEBUG=1;dotnet test` in your project. Note the process id it reports back to you.
Now pass that into vimspector when it asks you for the PID.

jpfeiffer16:
I don't think it would be to hard to extend this to run all test in a project once it is in.

nickspoons:
I'm trying this PR out with @jpfeiffer16's associated OmniSharp-vim PR and it works really well. In OmniSharp-vim we have easy access to the project metadata, such as executable build locations and project root locations, that are annoying to find manually to configure .vimspector.json files, so this PR is going to make a lot of things smoother and faster in that regard.

puremourning:
No worries. There’s no rush.

jpfeiffer16:
@puremourning . I've got a question for you.
So iIve added that test without the `lcd` as you requested, and it uncovered an interesting edge case.
When you launch the debugger, if your vim cwd is above the root of the dotnet solution file, then the debugger can't properly comunicate breakpoints and filepaths.
So, my question is this: Is this something we can just live with or should there be a way to pass the cwd into the debugger session in the `StartWithConfigurations(...)` function?


I've tried using the `sourceFileMap` property in the debug configuration, but I can't get it to work, since there is not really any part of the path to "replace" with the path relative to the vim cwd since the path that needs to be doctored is a subset of the path that needs to be resolved.

PR #479, opened 2021-10-15
Fix tracebacks when a channel connection fails.

puremourning:
Previously we ignored the result of 'start debug adapter' and this meant
that we crashed (vim traceback) when trying to initialise a null
connection. Solution: Bail out.

We also had a bug where if a channel-based connection failed, we always
said "Channel is already running" or something. This was because we
weren't clearing up the s:ch channel in that case. SOlution: clear it


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/479)

PR #490, opened 2021-11-17
Filter configurations by filetype

puremourning:
You can now specify `filetypes` in the `configuration` block as a list of vim filetypes the configuration applies to. When the list is specified for a configuration and no current buffer filetype is in the list, then the configuration is ignored.

This allows for better workflow when you have multiple types of file in your project. You can combine this with the 'default' and 'autoselect' entries to make default configurations for specific filetypes within your project.

Suggestion by @SUPERustam

### DIFF
--- a/README.md
+++ b/README.md
@@ -842,7 +842,8 @@ See [the reference guide][vimspector-ref-config-selection] for details.
 
 ### Get configurations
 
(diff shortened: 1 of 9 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/490)

PR #505, opened 2021-12-23
Make output visible for delve

### LINKED ISSUE #504: Go Delve dap `stdout` not visible
### Describe the bug

Thanks for solving #186, it works well! except for one thing: the `stdout` (and `stderr`) of the program I try to debug are visible nowhere. (I tried `VimspectorToggleLog` but there was nothing related to the output of the program).

It seems other projects are having the same issue: https://giters.com/mfussenegger/nvim-dap/issues/232
Aparrently `dlv dap` is printing stdout/stderr to its own process instead of redirecting it with DAP.
are you aware of that situtation, and if yes, do you have an idea for a workaround ?

### Minimal reproduciton

* Does your issue reproduce using `vim --clean -Nu /path/to/vimspector/support/minimal_vimrc` ? \[Yes]
* If you are using Neovim, does your issue reproduce using Vim? \[Yes]

> List of steps to reproduce:

> 1. Run `vim ---clean Nu /path/to/vimspector/support/minimal_vimrc`
> 2. Open any project that outputs something to stdout
> 3. Start debugging and observe an empty `console`

> Use the following Vimspector config file:

```json
{
  "configurations": {
    "run": {
      "adapter": "delve",
      "filetypes": ["go"],
      "variables": {
        "dlvFlags": "--check-go-version=false --log"
      },
      "configuration": {
        "request": "launch",
        "program": "${fileDirname}",
        "mode": "test"
      }
    }
  }
}
```

### Expected behaviour

The stdout should be printed out to the console view in debugging mode, but it isn't.

  
### Declaration

> Please complete the following declaration. If this declaration is not completed, your issue may be closed without comment.

* I have read and understood [CONTRIBUTING.md](https://github.com/puremourning/vimspector/blob/master/CONTRIBUTING.md) \[Yes]

puremourning:
Fixes #504 

There are two mechanisms implemented here - using a terminal buffer and using the Output window.

I'm actually not sure which is better. For now tty is the default for delve. This can be decided by using the following:

```
      "adapter": {
        "extends": "delve",
        "tty": true
      },

```

Setting "tty" to `false` will instead put the output in a new "Adapter" tab in the Output window.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/505)

puremourning:
I had to drop support for the output window because of some weird deadlock where the channel isn't closing properly on shutdown, although the job is stopped. I think the terminal version is better anyway.

puremourning (commented):
:lgtm:

> Reviewed 4 of 6 files at r1, 15 of 17 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/505#-:-MsAooL_7ksj5IXo7HMN:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @puremourning)

PR #506, opened 2021-12-30
Compact variables UI

### LINKED ISSUE #487: [Feature Request]: Truncate large arrays in variable window
I work a lot with large matrices and in the vimspector variables window, the array is shown completely which clutters the window a lot and makes it very uncomfortable to use it. 

See also the screenshot below. 

I would like the matrices in the variable window to be truncated and only show them when I expand the variable since then I want to see it. 
So maybe only show the first and last row or limit the amount of lines a variable preview can take. 

I have not considered any alternatives yet.

**Additional context**
Screenshot of variables window in vimspector showing a large matrix. 

The example uses python with debugpy. So I am not sure whether the problem exists in other programming languages or debug adapters as well. 

I hope my idea is clear and thanks for creating this great plugin.

puremourning:
This change removes the "type" from the variables and watches windows, replacing it with a hover popup that contains it. This is particularly useful for certain debug adapters like delve which include lots of repeated/duplicated data about the type of variables for some reason.

This also truncates the value so that we only display the first line of any multi-line values (again, the full value is visible in the hover popup).  Closes #487 (probably)

I'm changing the default here, which might annoy users, so there's an option `g:vimspector_variables_display_mode = 'full'` that restores the old behaviour. I may decide to remove it in future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1001,6 +1001,7 @@ autocmd SessionLoadPost * silent! VimspectorLoadSession
 * Use `<CR>`, or double-click with left mouse to expand/collapse (+, -).
 * Set the value of the variable with `<C-CR>` (control + `<CR>`) or
   `<leader><CR>` (if `modifyOtherKeys` doesn't work for you)
+* View the type of the variable via mouse hover.
 * When changing the stack frame the locals window updates.
 * While paused, hover to see values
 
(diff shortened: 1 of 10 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/506)

puremourning (commented):
:lgtm:

should really add a test for the hover-in-variables-window but who has the time?

> Reviewed 11 of 11 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/506#-:-MsBPxLc7A3maqLh0JCZ:b5gmd8c)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @puremourning)

PR #513, opened 2022-01-11
Mergify: Try to migrate away from strict mode, which was depricated

puremourning:


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/513)

Issue #514, opened 2022-01-13
[Feature Request]: Add .vim to the config search path


m-lima:
**Is your feature request related to a problem? Please describe.**
Vim configuration files are often stored in a `.vim` directory, e.g. `.vim/coc-setting.json` overrides project-specific CoC settings. This makes it easier to add to `.gitignore` IDE configuration directories like `.idea`, `.vscode`, `.vim`, etc.

Currently, vimspector searches up the path tree for root `.vimspector.json` and doesn't consider the `.vim` directory. Since vimspector is vim-specific, it would make sense and be cleaner to make it aware of this directory.

**Describe the solution you'd like**
When walking up the tree path, consider not only the `.vimspector.json`, but also `.vim/.vimspector.json` files.

**Describe alternatives you've considered**
Other alternative would be to make the file locations configurable.

puremourning:
seems related https://github.com/puremourning/vimspector/issues/495 and https://github.com/puremourning/vimspector/issues/380

I don't think that `$PROJECT_ROOT/.vim` is particularly canonical or de factor. I haven't see a precedent for that (other than you mention  coc, which meh) - are there other examples?  I initially found it confusing because I read `.vim` to mean `~/.vim` and though immediately that the solution was to use `g:vimspector_base_dir`.

It just doesn't seem that hard to add `.vimspector.*` to your `.gitignore`.

PR #515, opened 2022-01-13
Try .vim directory for config files

m-lima:
To address https://github.com/puremourning/vimspector/issues/514, this PR adds the **.vim** directory in the search path when walking up the path tree.

This allows users to include `.vim` in their `.gitignore` file for all Vim-related local settings and project-specific settings.
Therefore, it would be possible to add launch configurations, for example to Rust projects (which often require hand-written `program` variables), in a well-known "IDE" configuration folder, like what `coc-settings.json` does and other IDEs like `.idea` and `.vscode`


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/515)

puremourning (changes requested):
Thanks for sending a PR for this. That is a good way to sway my mind on whether I think it's a good idea or not :D

> rust... which often require hand-written program variables

Do they? I'd like to understand this better. 

> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/515#-:-MtIRDBS5g41k262PRNL:b-b7mmvf)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @m-lima and @puremourning)

*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/515#-MtIR6YQ6SWFmC8BE5tC:-MtIR6YR-fNP8Nkj-xCX:b82w85e) (no related file):*
Need to add tests

---
*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/515#-MtIR7Ue6GRlqYGry_2v:-MtIR7Ue6GRlqYGry_2w:bdnlxz7) (no related file):*
Need to update documentation

---

m-lima:
> Thanks for sending a PR for this. That is a good way to sway my mind on whether I think it's a good idea or not :D
> 
> > rust... which often require hand-written program variables
> 
> Do they? I'd like to understand this better.
> 
> > _[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/515#-:-MtIRDBS5g41k262PRNL:b-b7mmvf)_ status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @m-lima and @puremourning)
> 
> _[a discussion](https://reviewable.io/reviews/puremourning/vimspector/515#-MtIR6YQ6SWFmC8BE5tC:-MtIR6YR-fNP8Nkj-xCX:b82w85e) (no related file):_ Need to add tests
> 
> _[a discussion](https://reviewable.io/reviews/puremourning/vimspector/515#-MtIR7Ue6GRlqYGry_2v:-MtIR7Ue6GRlqYGry_2w:bdnlxz7) (no related file):_ Need to update documentation

The Rust comment was related to the fact that `cargo` produces randomized names for tests. There is a workaround by using `rust-analyzer` to populate the `${Executable}` and `${Args}` variables for a configuration.
Other than that, the solution that is more popular on the internet is to pre-run the tests, and check the name of the executable, so that it can be put into a configuration entry.

But, besides Rust, having a list of configurations with arguments prepopulated, for instance, for a project is a typical use of per-project `.vimspector.json` files. And allowing them to be organized in a `<project-dir>/.vim` directory like other IDEs do is a nice and cheap feature 😄 
At least, that's what I propose 😄 

And regarding the comment on the issue.. No, besides CoC, I cannot give you another example. Unless I go searching around for plugins 😛. At least my simple small set of installed plugins don't use it.
But I can see this being useful for plugins like `vim-rooter` or `vim-proj` to identify the vim root (often they use `.git` as a marker)

puremourning:
> The Rust comment was related to the fact that cargo produces randomized names for tests

Is there a way to run `cargo` and have it tell you the name of the test ? You can use a shell `variable` in that case to get the name by running some command into a vimspector variable.

puremourning:
ok so if you're using rust-analyser to provide the test input / replacement vars... do you still need to gitignore the file?

m-lima:
Me specifically, I have other configurations for conveniently running some entry points. Passing in args and env vars.
This is not a blocker for me ;)

The way I saw it was to allow this possibility, since the path is not customizable for vimspector.
It seems cheap enough, intuitive, and allows to group vim specific configuration like IDEs do, instead of multiple scattered files in the root.

I feel like you are leaning to close/reject this. If that is the case, just let me know and I won't bother updating the documentation 😅

puremourning:
nono, I think I'm ok with this change, just I also like to understand the pain points for reference. it's not the first time the cargo thing has come up and it's pretty unfortunate tbh.

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/515)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @m-lima and @puremourning)

*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/515#-MtIdRPAAaNuoN3cjO9R:-MtIdRPAAaNuoN3cjO9S:b-sz0i31) (no related file):*
Need to also consider updating the default locations for the session files `.vimspector.session`

---

m-lima:
> > _[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/515)_ status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @m-lima and @puremourning)
> 
> _[a discussion](https://reviewable.io/reviews/puremourning/vimspector/515#-MtIdRPAAaNuoN3cjO9R:-MtIdRPAAaNuoN3cjO9S:b-sz0i31) (no related file):_ Need to also consider updating the default locations for the session files `.vimspector.session`

I actually use this "feature" of this modification 🙂 
This PR does not change the default behavior, or assume that the user wants to store anything in a `.vim/` dir.
What is different is the finding of the session file. And this is what I use locally. I have my session files inside `.vim` and, with this change, the session is picked up.

I have to say, you built quite a stack around the development of this plugin! Really impressive stuff! There's plenty of automation!
And I think I hope I suspect I made the changes in the right places for the automation to pick it up and do docs, tests, and all that.
I ran the docker tester, but kept being prompted to install gadgets, even though I followed `CONTRIBUTING.md` and ran `./install_gadget.py --all`. So I'm pushing an update in the hope that the CI will do the right thing

puremourning:
> So I'm pushing an update in the hope that the CI will do the right thing

I've approved the CI run. I have to approve every run unfortunately due to Github rules/config about first-time contributors.

> I ran the docker tester, but kept being prompted to install gadgets

I'll check the instructions; they might be out of date.

You should be able to:

* `cd tests/manual`
* `./run`

Then within the container, you can do `./run_tests --base-dir test-base-docker --install` this should work.

e.g.


```
ben@BenMBP2021 manual % ./run
To run a command as administrator (user "root"), use "sudo <command>".
See "man sudo_root" for details.

dev@103a5220ea80:~$ pwd
/home/dev
dev@103a5220ea80:~$ ls
go  vimspector
dev@103a5220ea80:~$ cd vimspector/
dev@103a5220ea80:~/vimspector$ ./run_tests --base-dir test-base-docker --install --quiet --report messages
```

puremourning (changes requested):
> Reviewed 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/515)* status: all files reviewed, 2 unresolved discussions (waiting on @m-lima)

*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/515#-MtIdRPAAaNuoN3cjO9R:-Mu0wqmp3NhOylHWPNXW:bjc5gmn) (no related file):*
<details><summary><i>Previously, m-lima wrote…</i></summary><blockquote>

I actually use this "feature" of this modification 🙂 
This PR does not change the default behavior, or assume that the user wants to store anything in a `.vim/` dir.
What is different is the finding of the session file. And this is what I use locally. I have my session files inside `.vim` and, with this change, the session is picked up.

I have to say, you built quite a stack around the development of this plugin! Really impressive stuff! There's plenty of automation!
And I think I hope I suspect I made the changes in the right places for the automation to pick it up and do docs, tests, and all that.
I ran the docker tester, but kept being prompted to install gadgets, even though I followed `CONTRIBUTING.md` and ran `./install_gadget.py --all`. So I'm pushing an update in the hope that the CI will do the right thing

</blockquote></details>

OK great so I checked the code and the session file stuff uses `utils.PathToConfigFile` which is changed here. That allows the session file to exist in `.vim` but by default it still writes to `cwd`

That seems fine. I doubt many users will know about this quirk though, so it might be worth a note in the doc for session files. Optional, though because we don't currently document how it searches for session files, which I guess is an omission,

---
___
*[docs/configuration.md, line 419 at r2](https://reviewable.io/reviews/puremourning/vimspector/515#-Mu0xgeQ-hZqee_UKt7y:-Mu0xgeQ-hZqee_UKt7z:bz51d0v) ([raw file](https://github.com/puremourning/vimspector/blob/d8d5a01fd013beaad1673343f66180c421269617/docs/configuration.md#L419)):*
> ```Markdown
> * `<vimspector home>/configurations/<OS>/<filetype>/*.json`
> * `.vimspector.json` in the project source
> * `.vim/.vimspector.json` in the project source
> ```

we should also update the section for `.gadgets.json`. This is rarely used, but we've actually changed that too by changing `PathToConfigFile`

m-lima:
Updated the gadgets doc, added a few lines about how sessions are loaded (didn't go into much detail, just highlighting that `.vim/` can also house the path).
Lastly added the tests. Which is great that you asked for! :)
Rust doesn't care about the root.. Cargo simply walks up the tree to find a `Cargo.toml`, so this was not on my radar. But, by implementing the tests, I was able to see that the `workspace_root` is based off of the configuration file location.

From there, I had two possible approaches:
* Trim a workspace that is rooted on `.vim` up
* Extend `utils.py` to return two values: the file and the parent (very easy in `utils.py`)

However, the second approach would require lots more changes in `debug_session.py`.
I went for a more straightforward and simpler solution, simply walking up one level if the root is `.vim`.
My thoughts were that, the extra complexity and changes were not justified by the extremely rare (if even existent) corner case of a workspace root being `.vim` for any language. Not even VimL plugins root their sources under `.vim`. But really, I also didn't want to dump dozens of line changes for this.

But, I'm happy to go for the other approach if you prefer ;)

> My local tests behave differently from GHA. Maybe because I am using `aarch64`? ¯\\_(ツ)_/¯. In any case, I followed the output of GHA to guide through the tests, and it should be green now

puremourning (changes requested):
THanks, as mentioned below, I think I prefer something closer to the second approach. I'm less likely to break it.

> Reviewed 7 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/515#-:-Mu1na5_8vjDWHi_26Zs:bebu4qr)* status: all files reviewed, 3 unresolved discussions (waiting on @m-lima)

___
*[python3/vimspector/debug_session.py, line 186 at r4](https://reviewable.io/reviews/puremourning/vimspector/515#-Mu1mCIiENdR5xitUWwe:-Mu1mCIj-BZBWG6x9aKs:b-jyxtff) ([raw file](https://github.com/puremourning/vimspector/blob/9d4e840f809c346c6cac014b32ac3dfedba5b72f/python3/vimspector/debug_session.py#L186)):*
> ```Python
>     if launch_config_file:
>       self._workspace_root = os.path.dirname( launch_config_file )
>       if os.path.basename( self._workspace_root ) == '.vim':
> ```

yeah, this might be the least intrusive, but it's the most likely to break in future. As there's little chance that I will personally use this feature, it's just more likely that I will inadvertently break it or forget about it for some future change.

I would prefer that we changed `PathToConfigFile` to return a tuple of `file, root_dir` where `file` is the actual file path and `root_dir` is the "logical" directory in which file was found.

There are only like 3 places that we call `PathToConfigFile` so it's not that much change I think. I'd also be happy with a wrapper like `PathToConfigFile` just returns the file path, implemented interns of `PathToConfigFileWithRoot` or something. Dunno if that would actually be that helpful.

m-lima:
Closing this so it doesn't pollute your PR view
I guess it is not that important :)
Thanks for the reviews, though! 👍

PR #545, opened 2022-03-08
fix exception with key "line"

Maswor:
Working with Chrome debug, initially, the state of `pb` is:

```python
bp = {'state': 'ENABLED', 'line': 207, 'options': {}, 'sign_id': 3, 'server_bp': {'id': 1000, 'verified': False, 'message': 'Breakpoint set but not yet bound'}}
```
Screen Shot 2022-03-08 at 12 47 37 PM

Clearly, access `bp['server_bp']['line']` will throw an error (instead of return `None`) in Python 3. This PR make sure that we only access `line` key when it exists.

### DIFF
--- a/python3/vimspector/breakpoints.py
+++ b/python3/vimspector/breakpoints.py
@@ -278,7 +278,7 @@ class ProjectBreakpoints( object ):
       for bp in breakpoints:
         self._SignToLine( file_name, bp )
 
-        if 'server_bp' in bp:
+        if 'server_bp' in bp and 'line' in bp[ 'server_bp' ]:
           line = bp[ 'server_bp' ][ 'line' ]
           if bp[ 'server_bp' ][ 'verified' ]:
             state = 'VERIFIED'
@@ -819,7 +819,7 @@ class ProjectBreakpoints( object ):
           bp[ 'sign_id' ] = self._next_sign_id
           self._next_sign_id += 1
 
-        if 'server_bp' in bp:
+        if 'server_bp' in bp and 'line' in bp[ 'server_bp' ]:
           line = bp[ 'server_bp' ][ 'line' ]
           verified = bp[ 'server_bp' ][ 'verified' ]
         else:


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/545)

puremourning (changes requested):
Thanks for sending a PR!

Is this easy to reproduce? Would you be able to list steps so I can verify the fix ?

> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/545#-:-Mxf1ww25BdvFKE7J8PL:bj23qc4)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @Maswor and @puremourning)

___
*[python3/vimspector/breakpoints.py, line 281 at r1](https://reviewable.io/reviews/puremourning/vimspector/545#-Mxf0gVx6fahMK2p8mH0:-Mxf0gVy9uhvRkX948-y:ba22ema) ([raw file](https://github.com/puremourning/vimspector/blob/d5ed4a33c28384c2af7a40dcb3836d1371f3cf01/python3/vimspector/breakpoints.py#L281)):*
> ```Python
>         self._SignToLine( file_name, bp )
> 
>         if 'server_bp' in bp and 'line' in bp[ 'server_bp' ]:
> ```

I think this fix is the wrong one. What happens now is that if we have an unverified _server_ breakpoint, we render it as a _verified_ breakpoint on the user-selected line.

We should render it as a _unverified_ breakpoint at the user-requested line instead.

I think a better fix would be to put `line = bp{ 'server_bp' ][ 'line' ] if 'line' in ... else bp[ 'line' ]` (or better:

```
line = bp[ 'line' ]
if 'server_bp' in bp:
  if 'line' in bp[ 'server_bp' ]:
    line = bp[ 'server_bp' ][ 'line' ]
  ... set valid
```

Then we'll have to handle the "Jump To" command being a line of 0 too.
___
*[python3/vimspector/breakpoints.py, line 822 at r1](https://reviewable.io/reviews/puremourning/vimspector/545#-Mxf1NalDYenmUjeRzRF:-Mxf1NalDYenmUjeRzRG:b-gpr79z) ([raw file](https://github.com/puremourning/vimspector/blob/d5ed4a33c28384c2af7a40dcb3836d1371f3cf01/python3/vimspector/breakpoints.py#L822)):*
> ```Python
>           self._next_sign_id += 1
> 
>         if 'server_bp' in bp and 'line' in bp[ 'server_bp' ]:
> ```

as above, when we have a `server_bp` in the breakpoint object, we should use the `verified` that is within it. It's clearly possible for us to have "no line" at this time (when verified is falsy), so we'll need to handle that and _not_ place the sign.. I guess as the `bp[ 'line' ]` instead.

puremourning (commented):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/545)* status: all files reviewed, 2 unresolved discussions (waiting on @Maswor)

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/545)* status: all files reviewed, 2 unresolved discussions (waiting on @Maswor)

___
*[python3/vimspector/breakpoints.py, line 281 at r1](https://reviewable.io/reviews/puremourning/vimspector/545#-Mxf0gVx6fahMK2p8mH0:-Mxf2DRyBjLQcrpmCpQ8:bolw7yk) ([raw file](https://github.com/puremourning/vimspector/blob/d5ed4a33c28384c2af7a40dcb3836d1371f3cf01/python3/vimspector/breakpoints.py#L281)):*
ignore this...

> Then we'll have to handle the "Jump To" command being a line of 0 too.

puremourning:
Thanks again! I've pushed #548 to solve this.

PR #569, opened 2022-04-01
Re-use the log spave for the breakpoints window

puremourning:


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/569)

puremourning:
Doesn't re-use the breakpoints window space if you have the output window hidden then open breakpoints and then open output.

PR #627, opened 2022-07-20
Fix README typo

Alex-vZyl:
Just a small typo fix.  You can fix it yourself if you don't want to merge a full PR for it,


puremourning (commented):
Thanks for the RP!

The problem here is that TOC is autogenerated, so it gets broken every time it is updated. I'll probably have to fix the auto generator or change the section heading to use CPP rather than C++ or something sad...

> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/627#-:-N7QUnUG7yqgzZ8BBY8t:b-p5o5iy)* status: 0 of 1 files reviewed, all discussions resolved

Alex-vZyl:
Ah, I thought that might be the case...  Using CPP would definitely look better than just C ;)

PR #637, opened 2022-07-30
small gadgets installation fix

omercsp:
See commit messages for specific details

### DIFF
--- a/install_gadget.py
+++ b/install_gadget.py
@@ -156,7 +156,8 @@ installer.AbortIfSUperUser( args.sudo )
 
 vimspector_base = os.path.dirname( __file__ )
 if args.basedir:
-  vimspector_base = os.path.abspath( args.basedir )
+  vimspector_base = os.path.expanduser( args.basedir.strip() )
+  vimspector_base = os.path.abspath( vimspector_base )
 
 install.MakeInstallDirs( vimspector_base )
 installer.Configure( vimspector_base = vimspector_base,


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/637)

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/637)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @omercsp and @puremourning)

___
*[`.gitignore` line 26 at r1](https://reviewable.io/reviews/puremourning/vimspector/637#-N8Eem4H9mWJFXl6OpdI:-N8Eem4I0h1wkD66JWGK:b-2irbre) ([raw file](https://github.com/puremourning/vimspector/blob/6747e07014a3d11516fff66cdd89dbc1da0fb7e0/.gitignore#L26)):*
> ```ignore
> Session.vim
> .vimspector.session
> pyrightconfig.json
> ```

this doesn't seem relevant to this PR. Pls should only change one thing.

I'm also not sure if we really want to ignore this. I don't use pyright, but if I did, wouldn't I want the config in source control.
___
*[`install_gadget.py` line 159 at r1](https://reviewable.io/reviews/puremourning/vimspector/637#-N8Ef3L25U1Lp0jk3ryG:-N8Ef3L25U1Lp0jk3ryH:b-qqtmev) ([raw file](https://github.com/puremourning/vimspector/blob/6747e07014a3d11516fff66cdd89dbc1da0fb7e0/install_gadget.py#L159)):*
> ```Python
> vimspector_base = os.path.dirname( __file__ )
> if args.basedir:
>   vimspector_base = os.path.expanduser( args.basedir.strip() )
> ```

I think he problem with doing this is that vimspector _itself_ won't expand values like this.

So if you put `let g:vimspector_base_dir = '~/blah'` in your vimrc, now the installer migtht expand it, but vimspector won't find it.

Right?

Can you write on the PR description how you tested this change?

puremourning:
Thanks for sending a PR. can you please set the PR title to something that represents a summary of what is being changed? I'd also like to see a description of how this was tested and the exact use case, because on the face of it you would have to do `let g:vimspector_base_dir = expand( "~/blah" )` anyway

omercsp:
> Can you write on the PR description how you tested this change?

> this doesn't seem relevant to this PR. Pls should only change one thing.

> I think he problem with doing this is that vimspector _itself_ won't expand values like this.
> So if you put `let g:vimspector_base_dir = '~/blah'` in your vimrc, now the installer migtht expand it, but vimspector won't find it.
> Right?
I ran into this while trying to install through CLI the firefox-js adapter (The generated URL doesn't seem to work, BTW), and I use a none-default location for my gadgets.

My understanding of the code isn't that deep - but if installing a gadget from vim isn't translated to calling the install_gadgets.py with '--basedir' argument set to  'g:vimspector_base_dir' - than you are correct. A 'true' fix should include the unification of these two flows so the basdir manipulation will happen regardless of how the installation is invoked. This specific fix is just about the CLI. Let me know if you still want it and I'll update the PR with the single patch and some comments, or you want to have a full fix (which I don't think I'm going to do right now...) and close it.

puremourning:
> I ran into this while trying to install through CLI the firefox-js adapter (The generated URL doesn't seem to work, BTW), and I use a none-default location for my gadgets.

I didn't understand this, sorry... could you elaborate?

omercsp:
> I didn't understand this, sorry... could you elaborate?

This is a side note. The I tried to install the `vscode-firefox-debug` gadget, it failed,  and this is how I started digging the code to figure what went wrong and ran into that `basedir` problem from there.

But it's an unrelated problems. Vimspector fails (or at least - failed) to download the 'firefox' gadget for some issue with generated URL, that returns an error (same result with `curl`) - regardless to the `basedir` issue.

puremourning:
I'm still unsure how to reproduce any issue here, and the questions bout g:vimspector_base_dir remain. This PR seems abandoned, so closing. 

We can re-open if you can provide steps to repro the original issue and detailed regression testing notes.

Issue #633, opened 2022-07-25
[Feature Request]: PC/BP movements


dkaszews:
**Is your feature request related to a problem? Please describe.**
Vimspector supports jumping to BP or PC by clicking the breakpoint or frame respectively. This however requires moving over to a different window with either mouse or cursor. It also does not allow for advanced actions as using those movements in visual or operator-pending modes, e.g. to yank text between two breakpoints.

**Describe the solution you'd like**
Provide `<Plug>vimspector#JumpToPC`, `JumpToNextBP`, `JumpToPrevBP`.

**Describe alternatives you've considered**
Exposing functions `vimspector#GetPC` and `GetBPs` returning line number and buffer number or filename would allow user to implement such functionality themselves, with additional scripting possibilities. There are probably similar private functions at script scope, so exposing them should be easy. 

**Additional context**
More than happy to implement those myself, just need to confirm there is no easy way to do it already.

puremourning:
Please can you elaborate on your exact use case. You mention operator pending mappings, but I'm struggling to envisage the exact use case.

dkaszews:
OP-mode may be a stretch, I run into it couple times when the breakpoint was at a convenient place to yank text to, but I guess it was more of a coincidence.

I am mostly interested in normal mode mappings, so that I can jump between breakpoints and PC without leaving the code window. Most plugins add visual mode and operator-mode mappings just for consistency, but it is not required.

puremourning:
I could see something like `]B` and `[B` for next-breakpoint and previous-breakpoint (similar to diff mappings `[c`/`[c`) working. For jumping to the PC I tend to just use the up/down stack mappings that already exist.

dkaszews:
Yeah, exactly like that, will look into implementing it. Should be as easy as comparing list of breakpoints to cursor and calling `cursor`.

Moving stackframe up/down to jump to PC is a perfectly fine solution 😃

PR #639, opened 2022-08-01
Add JumpToNext/PreviousBreakpoint

dkaszews:
Tested automatically on vim and NeoVim using the written test methods.

Half of #633, will add `JumpToPC` in separate PR to keep them small (less implementation, more tests as need to let PC move about).

### DIFF
--- a/autoload/vimspector.vim
+++ b/autoload/vimspector.vim
@@ -469,6 +469,20 @@ function! vimspector#JumpToBreakpointViewBreakpoint() abort
   py3 _vimspector_session.JumpToBreakpointViewBreakpoint()
 endfunction
 
+function! vimspector#JumpToNextBreakpoint() abort
+  if !s:Enabled()
+    return
+  endif
+  py3 _vimspector_session.JumpToNextBreakpoint()
+endfunction
+
+function! vimspector#JumpToPreviousBreakpoint() abort
+  if !s:Enabled()
+    return
+  endif
+  py3 _vimspector_session.JumpToPreviousBreakpoint()
+endfunction
+
 function! vimspector#GetConfigurations() abort
   if !s:Enabled()
     return
--- a/README.md
+++ b/README.md
@@ -684,13 +684,15 @@ features to set your own mappings. To that end, Vimspector defines the following
 | `<Plug>VimspectorToggleBreakpoint`            | Toggle line breakpoint on the current line.                         | `vimspector#ToggleBreakpoint()`                                   |
 | `<Plug>VimspectorToggleConditionalBreakpoint` | Toggle conditional line breakpoint or logpoint on the current line. | `vimspector#ToggleBreakpoint( { trigger expr, hit count expr } )` |
 | `<Plug>VimspectorAddFunctionBreakpoint`       | Add a function breakpoint for the expression under cursor           | `vimspector#AddFunctionBreakpoint( '<cexpr>' )`                   |
-| `<Plug>VimspectorGoToCurrentLine`             | Reset the current program counter to the current line               | `vimspector#GoToCurrentLine()`                                        |
+| `<Plug>VimspectorGoToCurrentLine`             | Reset the current program counter to the current line               | `vimspector#GoToCurrentLine()`                                    |
 | `<Plug>VimspectorRunToCursor`                 | Run to Cursor                                                       | `vimspector#RunToCursor()`                                        |
 | `<Plug>VimspectorStepOver`                    | Step Over                                                           | `vimspector#StepOver()`                                           |
 | `<Plug>VimspectorStepInto`                    | Step Into                                                           | `vimspector#StepInto()`                                           |
 | `<Plug>VimspectorStepOut`                     | Step out of current function scope                                  | `vimspector#StepOut()`                                            |
 | `<Plug>VimspectorUpFrame`                     | Move up a frame in the current call stack                           | `vimspector#UpFrame()`                                            |
 | `<Plug>VimspectorDownFrame`                   | Move down a frame in the current call stack                         | `vimspector#DownFrame()`                                          |
+| `<Plug>VimspectorJumpToNextBreakpoint`        | Move Cursor to the next breakpoint in current file                  | `vimspector#JumpToNextBreakpoint()`                               |
+| `<Plug>VimspectorJumpToPreviousBreakpoint`    | Move Cursor to the previous breakpoint in current file              | `vimspector#JumpToPreviousBreakpoint()`                           |
 | `<Plug>VimspectorBalloonEval`                 | Evaluate expression under cursor (or visual) in popup               | *internal*                                                        |
 
 
--- a/plugin/vimspector.vim
+++ b/plugin/vimspector.vim
@@ -102,6 +102,12 @@ nnoremap <silent> <Plug>VimspectorUpFrame
 nnoremap <silent> <Plug>VimspectorDownFrame
       \ :<c-u>call <SID>SetRepeat( "\<Plug>VimspectorDownFrame" )<CR>
       \:<c-u>call vimspector#DownFrame()<CR>
+nnoremap <silent> <Plug>VimspectorJumpToNextBreakpoint
+      \ :<c-u>call <SID>SetRepeat( "\<Plug>VimspectorJumpToNextBreakpoint" )<CR>
+      \ :<c-u>call vimspector#JumpToNextBreakpoint()<CR>
+nnoremap <silent> <Plug>VimspectorJumpToPreviousBreakpoint
+      \ :<c-u>call <SID>SetRepeat( "\<Plug>VimspectorJumpToPreviousBreakpoint" )<CR>
+      \ :<c-u>call vimspector#JumpToPreviousBreakpoint()<CR>
 
 nnoremap <silent> <Plug>VimspectorBreakpoints
       \ :<c-u>call vimspector#ListBreakpoints()<CR>
--- a/python3/vimspector/breakpoints.py
+++ b/python3/vimspector/breakpoints.py
@@ -18,6 +18,7 @@ from collections import defaultdict
 import vim
 import os
 import logging
+import operator
 
 import json
 from vimspector import utils, signs, settings
@@ -292,27 +293,24 @@ class ProjectBreakpoints( object ):
 
     self.JumpToBreakpoint( bp )
 
-
-  def JumpToNextBreakpoint( self ):
+  def JumpToNextBreakpoint( self, reverse=False ):
     bps = self._breakpoints_view._breakpoint_list
     if not bps:
       return
 
     line = vim.current.window.cursor[ 0 ]
-    bp = next( ( bp for bp in bps if bp[ 'lnum' ] > line ), None )
+    comparator = operator.lt if reverse else operator.gt
+    bp = next(
+      ( bp
+        for bp in sorted( bps, key=operator.itemgetter( 'lnum' ), reverse=reverse )
+        if comparator( bp[ 'lnum' ], line ) ),
+      None )
+
     if bp:
       self.JumpToBreakpoint( bp )
 
   def JumpToPreviousBreakpoint( self ):
-    bps = self._breakpoints_view._breakpoint_list
-    if not bps:
-      return
-
-    line = vim.current.window.cursor[ 0 ]
-    bp = next( ( bp for bp in reversed(bps) if bp[ 'lnum' ] < line ), None )
-    if bp:
-      self.JumpToBreakpoint( bp )
-
+    self.JumpToNextBreakpoint( reverse=True )
 
   def ClearBreakpointViewBreakpoint( self ):
     bp = self._breakpoints_view.GetBreakpointForLine()
--- a/python3/vimspector/breakpoints.py
+++ b/python3/vimspector/breakpoints.py
@@ -266,14 +266,7 @@ class ProjectBreakpoints( object ):
     # so they are not touched
     self.UpdateUI()
 
-  def JumpToBreakpointViewBreakpoint( self ):
-    bp = self._breakpoints_view.GetBreakpointForLine()
-    if not bp:
-      return
-
-    if bp.get( 'type' ) != 'L':
-      return
-
+  def JumpToBreakpoint( self, bp ):
     success = int( vim.eval(
         f'win_gotoid( bufwinid( \'{ bp[ "filename" ] }\' ) )' ) )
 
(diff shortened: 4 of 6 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/639)

autoload/vimspector.vim
```diff
@@ -469,6 +469,20 @@ function! vimspector#JumpToBreakpointViewBreakpoint() abort
   py3 _vimspector_session.JumpToBreakpointViewBreakpoint()
 endfunction
 
+function! vimspector#JumpToNextBreakpoint() abort
```
dkaszews, line 472: New issue: This file is almost entirely composed of such forwarding functions. They could be generated from a list instead.

puremourning:
Nice thanks. 

One this worth testing is what happens when the server moves a breakpoint - ie put a bp on a var decl, start debugging and check where cursor goes. There is a distinction in the code between the user's requested position and the actual position the server used.

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/639)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @dkaszews and @puremourning)

___
*[`autoload/vimspector.vim` line 472 at r1](https://reviewable.io/reviews/puremourning/vimspector/639#-N8PSxGVQqsXVDKqaVN0-r1-472:-N8PvrPOABCtoRaU-CZU:b9jtrwn) ([raw file](https://github.com/puremourning/vimspector/blob/e205ae0c0506ddb1d5195d47899d29b071113fec/autoload/vimspector.vim#L472)):*
<details><summary><i>Previously, dkaszews wrote…</i></summary><blockquote>

New issue: This file is almost entirely composed of such forwarding functions. They could be generated from a list instead.
</blockquote></details>

Actually I prefer it this way. It’s explicit and makes it grepable.

dkaszews:
> > _[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/639)_ status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @dkaszews and @puremourning)
> 
> _[`autoload/vimspector.vim` line 472 at r1](https://reviewable.io/reviews/puremourning/vimspector/639#-N8PSxGVQqsXVDKqaVN0-r1-472:-N8PvrPOABCtoRaU-CZU:b9jtrwn) ([raw file](https://github.com/puremourning/vimspector/blob/e205ae0c0506ddb1d5195d47899d29b071113fec/autoload/vimspector.vim#L472)):_
> 
> Previously, dkaszews wrote…
> > New issue: This file is almost entirely composed of such forwarding functions. They could be generated from a list instead.
> 
> Actually I prefer it this way. It’s explicit and makes it grepable.

Understand the logic of not wanting fully generated functions, but what about slightly shortening it with a helper? Something like: `InvokePyIfEnabled` to make it a oneliner.

puremourning:
> > > _[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/639)_ status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @dkaszews and @puremourning)
> 
> > 
> 
> > _[`autoload/vimspector.vim` line 472 at r1](https://reviewable.io/reviews/puremourning/vimspector/639#-N8PSxGVQqsXVDKqaVN0-r1-472:-N8PvrPOABCtoRaU-CZU:b9jtrwn) ([raw file](https://github.com/puremourning/vimspector/blob/e205ae0c0506ddb1d5195d47899d29b071113fec/autoload/vimspector.vim#L472)):_
> 
> > 
> 
> > Previously, dkaszews wrote…
> 
> > > New issue: This file is almost entirely composed of such forwarding functions. They could be generated from a list instead.
> 
> > 
> 
> > Actually I prefer it this way. It’s explicit and makes it grepable.
> 
> 
> 
> Understand the logic of not wanting fully generated functions, but what about slightly shortening it with a helper? Something like: `InvokePyIfEnabled` to make it a oneliner. 

I just don't see the point. It's not like we add enough of these to warrant metaprogramming.

dkaszews:
Accidentally pushed changes to wrong branch, rolled back with `git push -f`

puremourning:
PythonLint CI is failing (you can run locally with `make check`)

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/639)* status: 0 of 4 files reviewed, 6 unresolved discussions (waiting on @dkaszews and @puremourning)

*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/639#-N8eDdd4Dle5zBOZd4Nz:-N8eDdd50zJJ6NQWQBJH:b-2gzq81) (no related file):*
We should add `<Plug>` mappings for these functions 

___
*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/639#-N8eDh-g3LWKElAUPgGf:-N8eDh-g3LWKElAUPgGg:b7re2pm) (no related file):*
We need to update `README.md`

___
*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/639#-N8eELP1B3QB6wC2exZP:-N8eELP1B3QB6wC2exZQ:b5xiq47) (no related file):*
need to add a test which does the jumping after a breakpoint is moved by the server to ensure we're using the right line number (i.e. the adjusted line number)

___
___
*[`python3/vimspector/breakpoints.py` line 270 at r2](https://reviewable.io/reviews/puremourning/vimspector/639#-N8eE-M974ePwgnfFee8:-N8eE-MAEVgRmMjP2O-w:b-22ovj5) ([raw file](https://github.com/puremourning/vimspector/blob/ebb38c130e93cc4db3b8146993f54b3dd8ffecd2/python3/vimspector/breakpoints.py#L270)):*
> ```Python
>     self.UpdateUI()
> 
>   def JumpToBreakpoint( self, bp ):
> ```

this should probably be a "internal" method now `_JumpToBreakpoint`

indeed it probably doesn't need to be a member anymore as it doesn't appear to use `self` anywhere, so we can promote this to a module scope function to make that distinction clear



_Code quote:_
```Python
JumpToBreakpoint
```
___
*[`python3/vimspector/breakpoints.py` line 306 at r2](https://reviewable.io/reviews/puremourning/vimspector/639#-N8eEdrRDV9SUxTSlIUW:-N8eEdrRDV9SUxTSlIUX:b-unulvw) ([raw file](https://github.com/puremourning/vimspector/blob/ebb38c130e93cc4db3b8146993f54b3dd8ffecd2/python3/vimspector/breakpoints.py#L306)):*
> ```Python
>       ( bp
>         for bp in sorted( bps, key=operator.itemgetter( 'lnum' ), reverse=reverse )
>         if comparator( bp[ 'lnum' ], line ) ),
> ```

I don't think that we can just use `bp[ 'line' ]` here as it will be "wrong" if the server moves/adjusts the line when verifying.

you can see elsewhere we have this sort of stuff:

```
        line = bp[ 'line' ]
        if 'server_bp' in bp:
          server_bp = bp[ 'server_bp' ]
          line = server_bp.get( 'line', line )
```

probably time for a `_GetBreakpointLine( bp )` rather than this

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/639)* status: 0 of 4 files reviewed, 7 unresolved discussions (waiting on @dkaszews and @puremourning)

___
*[`tests/breakpoints.test.vim` line 953 at r2](https://reviewable.io/reviews/puremourning/vimspector/639#-N8eF8SD5-PxxqlSYF8s:-N8eF8SD5-PxxqlSYF8t:bs6jch1) ([raw file](https://github.com/puremourning/vimspector/blob/ebb38c130e93cc4db3b8146993f54b3dd8ffecd2/tests/breakpoints.test.vim#L953)):*
> ```vim
>   lcd testdata/cpp/simple
>   edit simple.cpp
>   let main_win_id = win_getid()
> ```

this doesn't appear to be used



_Code quote:_
```vim
main_win_id
```

puremourning (commented):
> Reviewed 2 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/639)* status: all files reviewed, 7 unresolved discussions (waiting on @dkaszews and @puremourning)

dkaszews:
> I don't think that we can just use bp[ 'line' ] here as it will be "wrong" if the server moves/adjusts the line when verifying.

Will write a test for it, but I think it should work as it uses breakpoints window which I believe automatically updates.

dkaszews (commented):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/639)* status: all files reviewed, 6 unresolved discussions (waiting on @puremourning)

*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/639#-N8eELP1B3QB6wC2exZP:-N92jhD1-00CfpeCsdbT:bww4rec) (no related file):*

need to add a test which does the jumping after a breakpoint is moved by the server to ensure we're using the right line number (i.e. the adjusted line number)
</blockquote></details>

Added test, verified it works.

___
*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/639#-N8eDdd4Dle5zBOZd4Nz:-N92lvvp0V7_pIJWzpq9:bb6un1j) (no related file):*

We should add `<Plug>` mappings for these functions 
</blockquote></details>

Done

___
*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/639#-N8eDh-g3LWKElAUPgGf:-N92lxFW9Ea9PfAvmfec:b-i2kwd) (no related file):*

We need to update `README.md`
</blockquote></details>

Added to mappings table

___
___
*[`python3/vimspector/breakpoints.py` line 270 at r2](https://reviewable.io/reviews/puremourning/vimspector/639#-N8eE-M974ePwgnfFee8:-N92m1ep0yAcv7xLhA48:bqs0h67) ([raw file](https://github.com/puremourning/vimspector/blob/ebb38c130e93cc4db3b8146993f54b3dd8ffecd2/python3/vimspector/breakpoints.py#L270)):*

this should probably be a "internal" method now `_JumpToBreakpoint`

indeed it probably doesn't need to be a member anymore as it doesn't appear to use `self` anywhere, so we can promote this to a module scope function to make that distinction clear
</blockquote></details>

Moved to module scope, prefixed with `_`
___
*[`python3/vimspector/breakpoints.py` line 306 at r2](https://reviewable.io/reviews/puremourning/vimspector/639#-N8eEdrRDV9SUxTSlIUW:-N92jmInEZGghOnQTmsD:b-6lminm) ([raw file](https://github.com/puremourning/vimspector/blob/ebb38c130e93cc4db3b8146993f54b3dd8ffecd2/python3/vimspector/breakpoints.py#L306)):*

I don't think that we can just use `bp[ 'line' ]` here as it will be "wrong" if the server moves/adjusts the line when verifying.

you can see elsewhere we have this sort of stuff:

```
        line = bp[ 'line' ]
        if 'server_bp' in bp:
          server_bp = bp[ 'server_bp' ]
          line = server_bp.get( 'line', line )
```

probably time for a `_GetBreakpointLine( bp )` rather than this 
</blockquote></details>

Test shows it works, because the breakpoint is taken from automatically updated breakpoints window immediatelly before jump
___
*[`tests/breakpoints.test.vim` line 953 at r2](https://reviewable.io/reviews/puremourning/vimspector/639#-N8eF8SD5-PxxqlSYF8s:-N92k0u0DwzsJHSd71sq:ba3dfob) ([raw file](https://github.com/puremourning/vimspector/blob/ebb38c130e93cc4db3b8146993f54b3dd8ffecd2/tests/breakpoints.test.vim#L953)):*

this doesn't appear to be used
</blockquote></details>

Removed

puremourning (approved):
:lgtm:

Thanks!

> Reviewed 22 of 22 files at r4.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/639#-:-N92pVwI889TVYSAQoGW:b9benhj)* status: 6 of 23 files reviewed, 1 unresolved discussion (waiting on @puremourning)

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/639)* status: 6 of 23 files reviewed, 1 unresolved discussion (waiting on @dkaszews and @puremourning)

___
*[`tests/breakpoints.test.vim` line 1010 at r6](https://reviewable.io/reviews/puremourning/vimspector/639#-N92xHqo09FE0T4bcWxy:-N92xHqo09FE0T4bcWxz:b89y2d8) ([raw file](https://github.com/puremourning/vimspector/blob/3dffed554926a3cd46e91d5aea96778908b65d3e/tests/breakpoints.test.vim#L1010)):*
> ```vim
>   " First assert is needed to wait for launch to finish before moving cursor
>   call vimspector#Launch()
>   call vimspector#test#signs#AssertCursorIsAtLineInBuffer( 'simple.cpp', s:break_main_line, 1 )
> ```

80 characters max. Break at the `,`

ditto elsewhere



_Code quote:_
```vim
  call vimspector#test#signs#AssertCursorIsAtLineInBuffer( 'simple.cpp', s:break_main_line, 1 )
```

puremourning (commented):
> Reviewed 16 of 17 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/639)* status: all files reviewed, 1 unresolved discussion (waiting on @dkaszews)

dkaszews (commented):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/639)* status: 22 of 23 files reviewed, 1 unresolved discussion (waiting on @puremourning)

___
*[`tests/breakpoints.test.vim` line 1010 at r6](https://reviewable.io/reviews/puremourning/vimspector/639#-N92xHqo09FE0T4bcWxy:-N92ysrwBS6TbYWWqBiM:b-yjmi9w) ([raw file](https://github.com/puremourning/vimspector/blob/3dffed554926a3cd46e91d5aea96778908b65d3e/tests/breakpoints.test.vim#L1010)):*

80 characters max. Break at the `,`

ditto elsewhere
</blockquote></details>

Done, shouldn't it be checked by `VimscriptLint`?

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/639)* status: 22 of 23 files reviewed, all discussions resolved (waiting on @puremourning)

___
*[`tests/breakpoints.test.vim` line 1010 at r6](https://reviewable.io/reviews/puremourning/vimspector/639#-N92xHqo09FE0T4bcWxy:-N92zFXH6cSwMJ3Ftt58:b-aa7f9r) ([raw file](https://github.com/puremourning/vimspector/blob/3dffed554926a3cd46e91d5aea96778908b65d3e/tests/breakpoints.test.vim#L1010)):*
<details><summary><i>Previously, dkaszews (Dominik Kaszewski) wrote…</i></summary><blockquote>

Done, shouldn't it be checked by `VimscriptLint`?
</blockquote></details>

`vint` doesn't have this check. I tried to add it myself, but vint's AST seemed to be lacking the info to do this check

puremourning (approved):
:lgtm:

> Reviewed 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/639#-:-N92zPqv5k5_fNwQkMLY:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dkaszews)

PR #641, opened 2022-08-04
Fix test framework under NeoVim

### LINKED ISSUE #640: Cannot run tests under NeoVim
### Description

I know NeoVim support is experimental, but developing #639 I wanted to also test it aside from regular vim and noticed some very simple issues:

1. Unknown option: `--not-a-term` - from my observations can be omitted, neovim has similar option `--headless` but it does not render the window at all, which may break some tests 
1. Unknown functions: `ch_logfile`, `test_override`, `popup_close` - non-critical, can be omited if not exist
1. Unknown variable: `v:none` - use `v:null` supported by both vim and Neovim

Fixing those couple things allowed me to pass most tests on NeoVim. Some still fail on asserting current buffer or line for unclear reasons, but it's still better than nothing.

### Works in vim?

Only broken in Neovim

### Works with clean config?

Yes

### Sample project works?

No sample project for this filetype

### Minimal Reproduction

1. alias `vim` to `neovim` or modify `run_tests:RUN_VIM`
1. Add minimal test `tests/neovim.test.vim:
    ```
    function! SetUp() abort
      call vimspector#test#setup#SetUpWithMappings( v:null )
    endfunction

    function! Test_Can_Run_In_NeoVim()
      call assert_true( 1 )
    endfunction
    ```
1. `run_tests neovim.test.vim`

### Expected Behaviour

Test passes

### Actual Behaviour

Multiple errors in order of fixes:
1. `vim: Unknown option argument "--not-a-term"`
1. `E117: Unknown function: ch_logfile`
1. `E117: Unknown function: test_override`
1. `E121: Unknown variable: v:none`
1. `E117: Unknown function: popup_close`

### Additional information

_No response_

### Installation Type

Plugin Manager (git clone)

### Vimspector version

ce7ae8b

### Debug Info

_No response_

### Log file

_No response_

### Vim version

```Text
NVIM v0.6.1
Build type: Release
Lua 5.1
Compiled by team+vim@tracker.debian.org

Features: +acl +iconv +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info
```
```


### Python version

_No response_

### Neovim diagnostics

```Text
nvim: health#nvim#check
========================================================================
## Configuration
  - OK: no issues found

## Performance
  - OK: Build type: Release

## Remote Plugins
  - OK: Up to date

## terminal
  - INFO: key_backspace (kbs) terminfo entry: key_backspace=\177
  - INFO: key_dc (kdch1) terminfo entry: key_dc=\E[3~
  - INFO: $TERM_PROGRAM='tmux'
  - INFO: $SSH_TTY='/dev/pts/0'

## tmux
  - OK: escape-time: 0
  - INFO: Checking stuff
  - WARNING: focus-events is not enabled. 'autoread' may not work.
    - ADVICE:
      - (tmux 1.9+ only) Set focus-events in ~/.tmux.conf:
          set-option -g focus-events on
  - INFO: $TERM: tmux-256color
  - WARNING: Neither Tc nor RGB capability set. True colors are disabled. 'termguicolors' won't work properly.
    - ADVICE:
      - Put this in your ~/.tmux.conf and replace XXX by your $TERM outside of tmux:
          set-option -sa terminal-overrides ',XXX:RGB'
      - For older tmux versions use this instead:
          set-option -ga terminal-overrides ',XXX:Tc'

provider: health#provider#check
========================================================================
## Clipboard (optional)      
  - OK: Clipboard tool found: tmux      
      
## Python 2 provider (optional)      
  - WARNING: No Python executable found that can import neovim. Using the first available executable for diagnostics.      
  - ERROR: Python provider error:      
    - ADVICE:      
      - provider/pythonx: Could not load Python 2:      
          python2 not found in search path or not executable.      
          python2.7 not found in search path or not executable.    
          python2.6 not found in search path or not executable.      
          /usr/bin/python is Python 3.10 and cannot provide Python 2.      
  - INFO: Executable: Not found      
      
## Python 3 provider (optional)      
  - INFO: g:python3_host_prog is not set.  Searching for python3 in the environment.      
  - INFO: Multiple python3 executables found.  Set g:python3_host_prog to avoid surprises.
  - INFO: Executable: /usr/bin/python3
  - INFO: Other python executable: /bin/python3
  - INFO: Python version: 3.10.4
  - INFO: pynvim version: 0.4.2 (outdated; from /usr/lib/python3/dist-packages/neovim)
  - WARNING: Latest pynvim is NOT installed: 0.4.3

## Python virtualenv
  - OK: no $VIRTUAL_ENV

## Ruby provider (optional)
  - WARNING: ruby and gem must be in $PATH.
    - ADVICE:
      - Install Ruby and verify that ruby and gem commands work.

## Node.js provider (optional)
  - WARNING: node and npm (or yarn) must be in $PATH.
    - ADVICE:
      - Install Node.js and verify that node and npm (or yarn) commands work.

## Perl provider (optional)
  - ERROR: perl provider error:
    - ADVICE:
      - "Neovim::Ext" cpan module is not installed

vim.lsp: require("vim.lsp.health").check()
========================================================================
  - INFO: LSP log level : WARN
  - INFO: Log path: /home/dkaszews/.cache/nvim/lsp.log
  - INFO: Log size: 1 KB

vim.treesitter: require("vim.treesitter.health").check()
========================================================================
  - INFO: Runtime ABI version : 13

```
```


### Operating System

Linux Cuddles-Berry 5.15.0-1011-raspi #13-Ubuntu SMP PREEMPT Thu Jun 2 11:44:34 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux

### Declaration

- [X] I have read and understood [CONTRIBUTING.md](https://github.com/puremourning/vimspector/blob/master/CONTRIBUTING.md#diagnostics)
- [X] I have read and understood the [CODE_OF_CONDUCT.md](https://github.com/puremourning/vimspector/blob/master/CODE_OF_CONDUCT.md)

dkaszews:
Couple small changes to unblock running tests under NeoVim. Tested to work for multiple simple tests (some still don't work, usually asserting line or buffer name). Does not impact Vim as all changes are either compatible (`v:null`) or conditional for NeoVim.

Closes #640

### DIFF
--- a/README.md
+++ b/README.md
@@ -2323,6 +2323,29 @@ hi link jsonComment Comment
     Or google it.
 13. Can vimspector build my code before debugging it? Can I deploy it to a remote host before debugging it?
     No, not really. Vimspector is just a debugger, not a task system or build automation system - there are other tools for that. There is however a hack you can use - you can use a 'shell' variable to execute a command and just discard the output. Other options are discussed in [this issue](https://github.com/puremourning/vimspector/issues/227)
+14. It's annoying to manually type in the PID when attaching. Do you have a PID picker? There's no PID picker in vimspector at the moment, but you could write something and wrap `vimspector#LaunchWithSettings( { 'ThePID': the_pid_i_picked } )`. Alternatively, you could use a `shell` variable to guess the PID, like this (which runs `pgrep vim | sort | tail -1` to get the 'highest' PID of the command to be debugged (NOTE: this is for debugging Vim. replace with something appropriate to your actual use case. If this doesn't make sense to you, you might be better off just typing in the PID).
+
+```json
+    "Attach: max PID": {
+      "adapter": "CodeLLDB",
+      "variables": {
+        "pid": {
+          "shell": [
+            "/bin/bash",
+            "-c",
+            "pgrep vim | sort | tail -1"
+          ]
+        }
+      },
+      "configuration": {
+        "request": "attach",
+        "program": "${workspaceRoot}/src/vim",
+        "expressions": "native",
+        "stopOnEntry#json": "${StopOnEntry:true}",
+        "pid": "${pid}"
+      }
+    },
+```
 
 
 Example `g:vimspector_adapters` and `g:vimspector_configurations`:
--- a/autoload/vimspector.vim
+++ b/autoload/vimspector.vim
@@ -30,11 +30,11 @@ EOF
 endfunction
 
 
-" In vim, py3eval( 'None' ) returns v:none
+" In vim, py3eval( 'None' ) returns v:null
 " In neovim, py3eval( 'None' ) returns v:null
 "
 " Vim != Neovim. Go figure.
-let s:None = has( 'nvim' ) ? v:null : v:none
+let s:None = has( 'nvim' ) ? v:null : v:null
 
 function! s:Initialised() abort
   return exists( 's:enabled' )
--- a/autoload/vimspector.vim
+++ b/autoload/vimspector.vim
@@ -30,11 +30,11 @@ EOF
 endfunction
 
 
-" In vim, py3eval( 'None' ) returns v:null
+" In vim, py3eval( 'None' ) returns v:none
 " In neovim, py3eval( 'None' ) returns v:null
 "
 " Vim != Neovim. Go figure.
-let s:None = has( 'nvim' ) ? v:null : v:null
+let s:None = has( 'nvim' ) ? v:null : v:none
 
 function! s:Initialised() abort
   return exists( 's:enabled' )
--- a/autoload/vimspector/internal/job.vim
+++ b/autoload/vimspector/internal/job.vim
@@ -184,7 +184,7 @@ function! vimspector#internal#job#StartCommandWithLog( cmd, category ) abort
   if job_status( s:commands[ a:category ][ index ] ) !=# 'run'
     echom 'Unable to start job for ' . string( a:cmd )
     redraw
-    return v:none
+    return v:null
   endif
 
   return bufnr( buf )
--- a/autoload/vimspector/internal/job.vim
+++ b/autoload/vimspector/internal/job.vim
@@ -184,7 +184,7 @@ function! vimspector#internal#job#StartCommandWithLog( cmd, category ) abort
   if job_status( s:commands[ a:category ][ index ] ) !=# 'run'
     echom 'Unable to start job for ' . string( a:cmd )
     redraw
-    return v:null
+    return v:none
   endif
 
   return bufnr( buf )
--- a/doc/vimspector.txt
+++ b/doc/vimspector.txt
@@ -2651,6 +2651,36 @@ FAQ ~
    'shell' variable to execute a command and just discard the output. Other
    options are discussed in this issue [57]
 
+8. It's annoying to manually type in the PID when attaching. Do you have a
+   PID picker? There's no PID picker in vimspector at the moment, but you
+   could write something and wrap "vimspector#LaunchWithSettings( {
+   'ThePID': the_pid_i_picked } )". Alternatively, you could use a 'shell'
+   variable to guess the PID, like this (which runs 'pgrep vim | sort | tail
+   -1' to get the 'highest' PID of the command to be debugged (NOTE: this is
+   for debugging Vim. replace with something appropriate to your actual use
+   case. If this doesn't make sense to you, you might be better off just
+   typing in the PID).
+>
+  "Attach: max PID": {
+    "adapter": "CodeLLDB",
+    "variables": {
+      "pid": {
+        "shell": [
+          "/bin/bash",
+          "-c",
+          "pgrep vim | sort | tail -1"
+        ]
+      }
+    },
+    "configuration": {
+      "request": "attach",
+      "program": "${workspaceRoot}/src/vim",
+      "expressions": "native",
+      "stopOnEntry#json": "${StopOnEntry:true}",
+      "pid": "${pid}"
+    }
+  },
+<
 Example 'g:vimspector_adapters' and 'g:vimspector_configurations':
 >
   let g:vimspector_adapters = #{
--- a/run_tests
+++ b/run_tests
@@ -12,13 +12,7 @@ BASEDIR=$(AbsDirOf "$0")/${VIMSPECTOR_TEST_BASE:-test-base}
 INSTALL=0
 UPDATE=0
 INSTALLER_ARGS=''
-
 VIM_EXE='vim'
-IS_NVIM=$($VIM_EXE --version | grep -cm1 NVIM)
-VERSION_OPTIONS=$([ "$IS_NVIM" ] && echo '' || echo '--not-a-term')
-
-RUN_VIM="${VIM_EXE} -N --clean ${VERSION_OPTIONS}"
-RUN_TEST="${RUN_VIM} -S lib/run_test.vim"
 
 # 1 is stdout
 out_fd=1
@@ -48,6 +42,11 @@ while [ -n "$1" ]; do
       VIMSPECTOR_TEST_STDOUT=$1
       shift
       ;;
+    "--exe")
+      shift
+      VIM_EXE=$1
+      shift
+      ;;
     "--quiet")
       shift
       # send the output to /dev/null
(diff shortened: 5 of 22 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/641)

puremourning (changes requested):
> Reviewed 19 of 19 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/641)* status: all files reviewed, 5 unresolved discussions (waiting on @dkaszews)

___
*[`autoload/vimspector.vim` line 33 at r1](https://reviewable.io/reviews/puremourning/vimspector/641#-N8eFzaI5tly0UTJpJ8Y:-N8eFzaJCpsQVhxhoCLg:b6tux26) ([raw file](https://github.com/puremourning/vimspector/blob/95aa625ca2819f235660774ab9a9bc6946f84d12/autoload/vimspector.vim#L33)):*
> ```vim
> 
> 
> " In vim, py3eval( 'None' ) returns v:null
> ```

unless I am missing something, this hasn't changed and nor should this code?
it's already predicated on has( 'nvim' )
___
*[`autoload/vimspector/internal/job.vim` line 187 at r1](https://reviewable.io/reviews/puremourning/vimspector/641#-N8eG7ZSBuT2EtCdWNBK:-N8eG7ZTBI2sbXdpcMF9:b4xbdd4) ([raw file](https://github.com/puremourning/vimspector/blob/95aa625ca2819f235660774ab9a9bc6946f84d12/autoload/vimspector/internal/job.vim#L187)):*
> ```vim
>     echom 'Unable to start job for ' . string( a:cmd )
>     redraw
>     return v:null
> ```

this will never run on neovim, so why are we changing it?
___
*[`tests/minimal.test.vim` line 1 at r1](https://reviewable.io/reviews/puremourning/vimspector/641#-N8eGHMz8aLi4f8Jb6jF:-N8eGHMz8aLi4f8Jb6jG:b-3s5g53) ([raw file](https://github.com/puremourning/vimspector/blob/95aa625ca2819f235660774ab9a9bc6946f84d12/tests/minimal.test.vim#L1)):*
> ```vim
> " This test does nothing except check the test framework itself
> ```

I'm not sure we need to keep this around, do we?
___
*[`tests/variables.test.vim` line 669 at r1](https://reviewable.io/reviews/puremourning/vimspector/641#-N8eGPbSD5SdKUcAf9t2:-N8eGPbSD5SdKUcAf9t3:bmcavjp) ([raw file](https://github.com/puremourning/vimspector/blob/95aa625ca2819f235660774ab9a9bc6946f84d12/tests/variables.test.vim#L669)):*
> ```vim
> 
>   call WaitForAssert( {->
>         \   assert_notequal( v:null, g:vimspector_session_windows.eval )
> ```

this feels flaky. If the comment way above is true, then vim returns v:none and neopvim returns v:null, so we have to differentiate otherwise this check is no longer checking the right thing. In vim, if g:vimspector_session_windows is "v:none", then this check will pass (false-positive)


ditto everywhere else in this patch
___
*[`tests/lib/run_test.vim` line 50 at r1](https://reviewable.io/reviews/puremourning/vimspector/641#-N8eGiRh058AVr9XMyrG:-N8eGiRi4Z2SjoXQakTO:b-7c6ccj) ([raw file](https://github.com/puremourning/vimspector/blob/95aa625ca2819f235660774ab9a9bc6946f84d12/tests/lib/run_test.vim#L50)):*
> ```vim
> 
> if exists( '*ch_logfile' )
>   call ch_logfile( 'debuglog', 'w' )
> ```

is there an equivalent of this for neovim?

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/641)* status: all files reviewed, 6 unresolved discussions (waiting on @dkaszews)

*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/641#-N8eHA1yDq7V1NvXhcEU:-N8eHA1z0imRqkzzjZ9L:b5kyp16) (no related file):*
Thanks for this, it's somewhat fewer changes than I was expecting/fearing!

Given these changes do all the tests pass? I'm not sure that there's a huge amount value in changes that only make a few of the tests work, because the state is still "not working" if any one test fails. If we can make them all work, and run in CI then great.

___

puremourning:
Tests are failing on vim

dkaszews:
I'll need to check it later. In general I just run a macro to change all `v:none` to `v:null` because from what I understand they are just aliases in vim, but I may be wrong. That's also why there may be changes in code already under `if !has( 'nvim' )` - it was easier and faster to just make sure there are no `v:none` in entire repo.

The `minimal.test.vim` can be removed, it was just for me to find a bare minimum.

> is there an equivalent of this for neovim

Unfortunately not: https://github.com/neovim/neovim/issues/9010 but I can try emulating it by copying the log file from home to appropriate location after every test.

puremourning:
In vim:

* `echo py3eval( 'None' ) is v:none` prints `1`
* `echo py3eval( 'None' ) is v:null` prints `0`
* `echo py3eval( 'None' ) == v:none` prints `1`:
* `echo py3eval( 'None' ) == v:null` prints `0`
* `echo v:null == v:none` prints `0`

puremourning:
No need to bother with the ch_log for neovim. if it's not there, we can live without it. the test fwd already scoops up lots of logs.

dkaszews:
Oh, that is really stupid. In vim, `v:null` and `v:none` are not equal, but both are of type `v:t_none`. In Neovim, `v:none` does not exist, and `v:null`'s type is `v:null`. So every `x == v:none` would have to be changed to `type(x) == type(x:null)`. Or better, encapsulate it into a `isNull` (because `is v:null` still does not work in vim). I'll take a look at it and try to figure out the best solution.

dkaszews (commented):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/641)* status: 13 of 20 files reviewed, 6 unresolved discussions (waiting on @dkaszews and @puremourning)

___
*[`autoload/vimspector.vim` line 33 at r1](https://reviewable.io/reviews/puremourning/vimspector/641#-N8eFzaI5tly0UTJpJ8Y:-N8jU3ap7jeWJMlNO6sW:b4re4w8) ([raw file](https://github.com/puremourning/vimspector/blob/95aa625ca2819f235660774ab9a9bc6946f84d12/autoload/vimspector.vim#L33)):*

unless I am missing something, this hasn't changed and nor should this code?
it's already predicated on has( 'nvim' )
</blockquote></details>

Accidental change, reverted
___
*[`autoload/vimspector/internal/job.vim` line 187 at r1](https://reviewable.io/reviews/puremourning/vimspector/641#-N8eG7ZSBuT2EtCdWNBK:-N8jUCQ8B_VVmm5LqMW1:b4re4w8) ([raw file](https://github.com/puremourning/vimspector/blob/95aa625ca2819f235660774ab9a9bc6946f84d12/autoload/vimspector/internal/job.vim#L187)):*

this will never run on neovim, so why are we changing it?
</blockquote></details>

Accidental change, reverted
___
*[`tests/minimal.test.vim` line 1 at r1](https://reviewable.io/reviews/puremourning/vimspector/641#-N8eGHMz8aLi4f8Jb6jF:-N8jUG4G9Co9Vz84ezJh:ba3dfob) ([raw file](https://github.com/puremourning/vimspector/blob/95aa625ca2819f235660774ab9a9bc6946f84d12/tests/minimal.test.vim#L1)):*

I'm not sure we need to keep this around, do we?
</blockquote></details>

Removed
___
*[`tests/variables.test.vim` line 669 at r1](https://reviewable.io/reviews/puremourning/vimspector/641#-N8eGPbSD5SdKUcAf9t2:-N8jUHki14PT_k96i9oT:b-1pq0fr) ([raw file](https://github.com/puremourning/vimspector/blob/95aa625ca2819f235660774ab9a9bc6946f84d12/tests/variables.test.vim#L669)):*

this feels flaky. If the comment way above is true, then vim returns v:none and neopvim returns v:null, so we have to differentiate otherwise this check is no longer checking the right thing. In vim, if g:vimspector_session_windows is "v:none", then this check will pass (false-positive)


ditto everywhere else in this patch
</blockquote></details>

Added helper `Assert(Not)Null` which should work correctly on both vim and neovim

dkaszews (commented):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/641)* status: 13 of 20 files reviewed, 6 unresolved discussions (waiting on @dkaszews and @puremourning)

*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/641#-N8eHA1yDq7V1NvXhcEU:-N92bvMI6GjM8KCW24qu:b7y2lvq) (no related file):*

Thanks for this, it's somewhat less than I was expecting/fearing.

Given these changes do all the tests pass? I'm not sure that there's a huge amount value in changes that only make a few of the tests work, because the state is still "not working" if any one test fails. If we can make them all work, and run in CI then great.
</blockquote></details>

From discussion on Gitter: many tests still fail, agreed to handle in separate PR

___

dkaszews (commented):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/641)* status: 13 of 20 files reviewed, 6 unresolved discussions (waiting on @puremourning)

___
*[`tests/lib/run_test.vim` line 50 at r1](https://reviewable.io/reviews/puremourning/vimspector/641#-N8eGiRh058AVr9XMyrG:-N92c8lVErJJybUyIgtV:bwrill2) ([raw file](https://github.com/puremourning/vimspector/blob/95aa625ca2819f235660774ab9a9bc6946f84d12/tests/lib/run_test.vim#L50)):*

is there an equivalent of this for neovim?
</blockquote></details>

Answered outside thread: no equivalent, could be emulated by copying files manually but not worth effort since most valuable logs already work

puremourning (approved):
Thanks!

:lgtm: 

> Reviewed 7 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/641#-:-N8eGvAQ3pMpIWRtKvnj:biwhqwx)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @dkaszews)

PR #648, opened 2022-08-12
Add JumpToProgramCounter

### LINKED ISSUE #633: [Feature Request]: PC/BP movements

dkaszews:
Closes #633

### DIFF
--- a/README.md
+++ b/README.md
@@ -729,6 +729,8 @@ let g:vimspector_enable_mappings = 'VISUAL_STUDIO'
 | `Shift F5`      | `<Plug>VimspectorStop`                  | Stop debugging.
 | `Ctrl Shift F5` | `<Plug>VimspectorRestart`               | Restart debugging with the same configuration.
 | `F6`            | `<Plug>VimspectorPause`                 | Pause debuggee.
+| `F8`            | `<Plug>VimspectorJumpToNextBreakpoint`  | Jump to next breakpoint in the current file.
+| `Shift F8`      | `<Plug>VimspectorJumpToPreviousBreakpoint` | Jump to previous breakpoint in the current file.
 | `F9`            | `<Plug>VimspectorToggleBreakpoint`      | Toggle line breakpoint on the current line.
 | `Shift F9`      | `<Plug>VimspectorAddFunctionBreakpoint` | Add a function breakpoint for the expression under cursor
 | `F10`           | `<Plug>VimspectorStepOver`              | Step Over
--- a/README.md
+++ b/README.md
@@ -693,6 +693,7 @@ features to set your own mappings. To that end, Vimspector defines the following
 | `<Plug>VimspectorDownFrame`                   | Move down a frame in the current call stack                         | `vimspector#DownFrame()`                                          |
 | `<Plug>VimspectorJumpToNextBreakpoint`        | Move Cursor to the next breakpoint in current file                  | `vimspector#JumpToNextBreakpoint()`                               |
 | `<Plug>VimspectorJumpToPreviousBreakpoint`    | Move Cursor to the previous breakpoint in current file              | `vimspector#JumpToPreviousBreakpoint()`                           |
+| `<Plug>VimspectorJumpToProgramCounter`        | Move Cursor to the program counter in the current frame             | `vimspector#JumpToProgramCounter()`                               |
 | `<Plug>VimspectorBalloonEval`                 | Evaluate expression under cursor (or visual) in popup               | *internal*                                                        |
 
 
--- a/autoload/vimspector.vim
+++ b/autoload/vimspector.vim
@@ -483,6 +483,13 @@ function! vimspector#JumpToPreviousBreakpoint() abort
   py3 _vimspector_session.JumpToPreviousBreakpoint()
 endfunction
 
+function! vimspector#JumpToProgramCounter() abort
+  if !s:Enabled()
+    return
+  endif
+  py3 _vimspector_session.JumpToProgramCounter()
+endfunction
+
 function! vimspector#GetConfigurations() abort
   if !s:Enabled()
     return
--- a/plugin/vimspector.vim
+++ b/plugin/vimspector.vim
@@ -119,6 +119,8 @@ if s:mappings ==# 'VISUAL_STUDIO'
   nmap <S-F5>       <Plug>VimspectorStop
   nmap <C-S-F5>     <Plug>VimspectorRestart
   nmap <F6>         <Plug>VimspectorPause
+  nmap <F8>         <Plug>VimspectorJumpToNextBreakpoint
+  nmap <S-F8>       <Plug>VimspectorJumpToPreviousBreakpoint
   nmap <F9>         <Plug>VimspectorToggleBreakpoint
   nmap <S-F9>       <Plug>VimspectorAddFunctionBreakpoint
   nmap <F10>        <Plug>VimspectorStepOver
--- a/plugin/vimspector.vim
+++ b/plugin/vimspector.vim
@@ -108,6 +108,8 @@ nnoremap <silent> <Plug>VimspectorJumpToNextBreakpoint
 nnoremap <silent> <Plug>VimspectorJumpToPreviousBreakpoint
       \ :<c-u>call <SID>SetRepeat( "\<Plug>VimspectorJumpToPreviousBreakpoint" )<CR>
       \ :<c-u>call vimspector#JumpToPreviousBreakpoint()<CR>
+nnoremap <silent> <Plug>VimspectorJumpToProgramCounter
+      \ :<c-u>call vimspector#JumpToProgramCounter()<CR>
 
 nnoremap <silent> <Plug>VimspectorBreakpoints
       \ :<c-u>call vimspector#ListBreakpoints()<CR>
--- a/python3/vimspector/debug_session.py
+++ b/python3/vimspector/debug_session.py
@@ -1773,6 +1773,9 @@ class DebugSession( object ):
   def JumpToPreviousBreakpoint( self ):
     self._breakpoints.JumpToPreviousBreakpoint()
 
+  def JumpToProgramCounter( self ):
+    self._stackTraceView.JumpToProgramCounter()
+
   def ToggleBreakpoint( self, options ):
     return self._breakpoints.ToggleBreakpoint( options )
 
--- a/python3/vimspector/stack_trace.py
+++ b/python3/vimspector/stack_trace.py
@@ -419,6 +419,14 @@ class StackTraceView( object ):
       self._JumpToFrame( frame, 'down' )
 
 
+  def JumpToProgramCounter( self ):
+    frame = self._GetFrameOffset( 0 )
+    if not frame:
+      utils.UserMessage( 'No current stack frame' )
+    else:
+      self._JumpToFrame( frame, 'jump' )
+
+
   def AnyThreadsRunning( self ):
     for thread in self._threads:
       if thread.state != Thread.TERMINATED:
--- a/tests/stack_trace.test.vim
+++ b/tests/stack_trace.test.vim
@@ -585,10 +585,12 @@ function! Test_JumpToProgramCounter() abort
   call vimspector#StepInto()
   call TestJumpToPCAux( break_foo_line )
 
-  edit 'struct.cpp'
+  edit struct.cpp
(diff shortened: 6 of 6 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/648)

puremourning:
> Should we add a default, opt-out mapping? 

Vimspector policy is not to change _any_ default mappings unless requested by the user.

From README:

> By default, vimspector does not change any of your mappings. Mappings are very personal and so you should work out what you like and use vim's powerful mapping features to set your own mappings. 

We _can_ change the mappings set by g:ycm_enable_mappings = "HUMAN" and _should_ change the "VISUAL_STUDIO" set if there are VS mappings that are equivalent to ones we're adding support for (though I don't use VS so I don't know these intimately)

dkaszews (commented):
From what I can see, Visual Studio has following mappings (I use it at work but am on vacation now):
* Display next statement - `Alt+Num*`
* Next location - `F8`
* Previous location - `Shift+F8`

None are perfect match, as I believe "display next statement" opens a popup instead of moving the cursor, while "next/previous location" depends on the window focused, so would only work if you focus breakpoints window. Still, I propose the following:

Visual Studio:
* Next (previous) breakpoint - `(Shift+)F8`
* No binding for jump to cursor - the VS one is not very "vimy", may conflict with regular `*` which is "find next under cursor"

Human:
* Next (previous) breakpoint - `(Shift+)F7` - `F8` already taken
* Jump to cursor - `F2` - only F-key left, on Windows means "rename" so kinda matches editing current thing

> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/648#-:-N9GmzER5gtGdjzGOAZb:byev40r)* status: 0 of 5 files reviewed, all discussions resolved

puremourning:
> Human:

> Next (previous) breakpoint - (Shift+)F7 - F8 already taken

one of the goals of human mode is to avoid things like shifted-f-keys which rarely work in termcap and require that you have too many hands to be comfortable.

> Jump to cursor - F2 - only F-key left, on Windows means "rename" so kinda matches editing current thing

ehhhhh, I'm not sure this feels right.


I think I'm concluding that we should probably not have default mappings for these things. It's just really hard to pick something that doesn't conflict and feels like it's in the right place.

I'm happy with the VS mappings you propose based on what you said. For human mode I think we can just let users pick something and map it themselves. I would prefer that it was a little painful to set up than accidentally breaks someone's muscle memory, while I kinda want to keep the F-keys free for other things (like reverse-step and stuff)

dkaszews (commented):
Sure, agreed that those "human" ones would be there just to be there. I personally will use `]b`, `[b` and `]p` as they fit most with vim and other plugins I use. So would be weird to propose defaults I don't like myself. 

> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/648#-:-N9I7kEo4BKe10s9oXV7:b-yka86k)* status: 0 of 5 files reviewed, all discussions resolved

dkaszews (commented):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/648)* status: 0 of 6 files reviewed, 1 unresolved discussion

___
*[`tests/stack_trace.test.vim` line 563 at r2](https://reviewable.io/reviews/puremourning/vimspector/648#-N9XGZZT-UHHi_PYqILm:-N9XGZZT-UHHi_PYqILn:bbnaamk) ([raw file](https://github.com/puremourning/vimspector/blob/4fecc83002b6062b0b4995cb0c28c83657d512a3/tests/stack_trace.test.vim#L563)):*
> ```vim
> endfunction
> 
> function! Test_JumpToProgramCounter() abort
> ```

Unsure whether this is a good file to put the test in, I figured it makes most sense from implementation perspective because the test is about `Up/DownFrame` which is implemented the same way.

puremourning (changes requested):
tests are failing

> Dismissed @dkaszews from a discussion.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/648#-:-N9XPHAs7ek1WCNxDSBj:b-smz0ks)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @dkaszews and @puremourning)

___
*[`tests/stack_trace.test.vim` line 563 at r2](https://reviewable.io/reviews/puremourning/vimspector/648#-N9XGZZT-UHHi_PYqILm:-N9XOuMe0mSRIrHqlxRg:b-c761qb) ([raw file](https://github.com/puremourning/vimspector/blob/4fecc83002b6062b0b4995cb0c28c83657d512a3/tests/stack_trace.test.vim#L563)):*
<details><summary><i>Previously, dkaszews (Dominik Kaszewski) wrote…</i></summary><blockquote>

Unsure whether this is a good file to put the test in, I figured it makes most sense from implementation perspective because the test is about `Up/DownFrame` which is implemented the same way.
</blockquote></details>

yep, it's fine here, thanks
___
*[`tests/stack_trace.test.vim` line 576 at r3](https://reviewable.io/reviews/puremourning/vimspector/648#-N9XPvG91he9y8CFixHz:-N9XPvG91he9y8CFixI-:bfw41bp) ([raw file](https://github.com/puremourning/vimspector/blob/389573f5678b003cde3f79f1196756d5f26e6659/tests/stack_trace.test.vim#L576)):*
> ```vim
>                                                          \ 1 )
> 
>   function! TestJumpToPCAux( line ) closure abort
> ```

make sure to add a `delfunc` for this at the end of this test
___
*[`tests/stack_trace.test.vim` line 590 at r3](https://reviewable.io/reviews/puremourning/vimspector/648#-N9XPnMdER9djfrgjwnt:-N9XPnMdER9djfrgjwnu:bnmhcgk) ([raw file](https://github.com/puremourning/vimspector/blob/389573f5678b003cde3f79f1196756d5f26e6659/tests/stack_trace.test.vim#L590)):*
> ```vim
>   call TestJumpToPCAux( 17 )
> 
>   edit 'struct.cpp'
> ```

very weird to have a filename with quotes in it, which I'm pretty sure this is trying to open `'struct.cpp'` not `struct.cpp`

puremourning (commented):
> Reviewed 3 of 5 files at r1, 2 of 3 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/648)* status: all files reviewed, 2 unresolved discussions (waiting on @dkaszews)

dkaszews (commented):
Trying to fix it, can you please help? I cannot repro it on my machine, and CI logs are unhelpful to me:

```
%FAIL: stack_trace.test.vim FAILED - see /__w/vimspector/vimspector/tests/logs/stack_trace.test.vim
%MESSAGES

From /__w/vimspector/vimspector/tests/stack_trace.test.vim:
Executing Test_JumpToProgramCounter()
Executing Test_Multiple_Threads_Continue()
Executing Test_Multiple_Threads_Step()
Executing Test_UpDownStack()
Executing abort
Wrote log for failed test: /__w/vimspector/vimspector/tests/stack_trace.test.vim_abort_vimspector.log.testlog
Wrote log for failed test: /__w/vimspector/vimspector/tests/stack_trace.test.vim_abort_vimspector.log.testlog
Executed 5 tests
1 FAILED:
Found errors in /__w/vimspector/vimspector/tests/stack_trace.test.vim:abort:
Caught exception in abort: Vim(call):E107: Missing parentheses: abort @ /__w/vimspector/vimspector/tests/stack_trace.test.vim:function RunTheTest, line 71
```

> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/648#-:-N9XRONR8dpcVyzzSDuS:by7tojg)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on @puremourning)

___
*[`tests/stack_trace.test.vim` line 576 at r3](https://reviewable.io/reviews/puremourning/vimspector/648#-N9XPvG91he9y8CFixHz:-N9XRe6R4mxa9zsF3eD2:b-4t4y44) ([raw file](https://github.com/puremourning/vimspector/blob/389573f5678b003cde3f79f1196756d5f26e6659/tests/stack_trace.test.vim#L576)):*

make sure to add a `delfunc` for this at the end of this test
</blockquote></details>

I thought functions within functions have `l:` scope by default, done.
___
*[`tests/stack_trace.test.vim` line 590 at r3](https://reviewable.io/reviews/puremourning/vimspector/648#-N9XPnMdER9djfrgjwnt:-N9XRmcI8P6a1u8BMS9_:bjuot0h) ([raw file](https://github.com/puremourning/vimspector/blob/389573f5678b003cde3f79f1196756d5f26e6659/tests/stack_trace.test.vim#L590)):*

very weird to have a filename with quotes in it, which I'm pretty sure this is trying to open `'struct.cpp'` not `struct.cpp`
</blockquote></details>

Yeah, it was just me being confused between commands and functions again 😄 . Fixed.

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/648)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @dkaszews and @puremourning)

___
*[`tests/stack_trace.test.vim` line 576 at r3](https://reviewable.io/reviews/puremourning/vimspector/648#-N9XPvG91he9y8CFixHz:-N9XVbIg48mNZG2RiK4U:b8dk479) ([raw file](https://github.com/puremourning/vimspector/blob/389573f5678b003cde3f79f1196756d5f26e6659/tests/stack_trace.test.vim#L576)):*
<details><summary><i>Previously, dkaszews (Dominik Kaszewski) wrote…</i></summary><blockquote>

I thought functions within functions have `l:` scope by default, done.
</blockquote></details>

nope, there's no such thing... all functions in vim are global

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/648)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on @dkaszews and @puremourning)

___
*[`tests/stack_trace.test.vim` line 563 at r4](https://reviewable.io/reviews/puremourning/vimspector/648#-N9XXE4QCkVOV-mjvwE6:-N9XXE4QCkVOV-mjvwE7:b1699at) ([raw file](https://github.com/puremourning/vimspector/blob/315d91b1ce95602277a5ae8786997c3c9be370ef/tests/stack_trace.test.vim#L563)):*
> ```vim
> endfunction
> 
> function! Test_JumpToProgramCounter() abort
> ```

remove `abort` from here, I think that's what's breaking CI

the test framework uses a very naive way to discover test functions: 

```
redir @q
silent function /^Test_
redir END
let s:tests = split(substitute(@q, 'function \(\k*()\)', '\1', 'g'))`
```

and well, it probably found this and doesn't correct parse the trailing `abort`

But more importantly, we actually need it to keep going on error to run the cleanup code at the end. We don't wrap the whole body in `try/finally` because that would be ugly.

dkaszews (commented):
I think I figured it out - marked my test as `abort`, which confused the test framework into trying to call `abort` as if it was a test function.

> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/648#-:-N9XXCA590mbwWmTY37_:b-bwry0o)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on @dkaszews and @puremourning)

___
*[`tests/stack_trace.test.vim` line 576 at r3](https://reviewable.io/reviews/puremourning/vimspector/648#-N9XPvG91he9y8CFixHz:-N9XXQoaDkaKp0ESUsXU:b-896fix) ([raw file](https://github.com/puremourning/vimspector/blob/389573f5678b003cde3f79f1196756d5f26e6659/tests/stack_trace.test.vim#L576)):*

nope, there's no such thing... all functions in vim are global
</blockquote></details>

Done.

PR #664, opened 2022-09-01
Enable PID picker injection

Aster89:
Fundamentally, this is the code you sketched.

I'm not sure how to test it. I guess by inspection we can be sure it doesn't alter the current behavior.

Whether a customization actually works I guess is another story. But that's maybe ok, as in, the user has to make sure to write an appropriate function to make it work.

Anyway, I have this in my vimrc to make use of it:
```
let g:vimspector_custom_process_picker = 'PickProcess()'

function! PickProcess() abort
  return str2nr(split(fzf#run({'source': 'ps -e'})[0])[0])
endfunction
```
which I think could go in an example in the doc.

If you are happy with it, I can do it.

TODO:

* [ ] docs
* [ ] tests for attachment

### DIFF
--- a/python3/vimspector/debug_session.py
+++ b/python3/vimspector/debug_session.py
@@ -435,6 +435,7 @@ class DebugSession( object ):
       # working directory.
       'cwd': os.getcwd,
       'unusedLocalPort': utils.GetUnusedLocalPort,
+      'selectProcess': _SelectProcess,
     }
 
     # Pretend that vars passed to the launch command were typed in by the user
@@ -1596,7 +1597,7 @@ class DebugSession( object ):
       if attach_config[ 'pidSelect' ] == 'ask':
         prop = attach_config[ 'pidProperty' ]
         if prop not in launch_config:
-          pid = utils.AskForInput( 'Enter PID to attach to: ' )
+          pid = _SelectProcess()
           if pid is None:
             return
           launch_config[ prop ] = pid
@@ -2204,3 +2205,10 @@ def PathsToAllConfigFiles( vimspector_base, current_file, filetypes ):
 
   yield utils.PathToConfigFile( '.vimspector.json',
                                 os.path.dirname( current_file ) )
+
+
+def _SelectProcess():
+  custom_picker = settings.Get( 'custom_process_picker_expr' )
+  if custom_picker:
+    return int( vim.eval( custom_picker ) )
+  return int( utils.AskForInput( 'Enter Process ID: ' ) )
--- a/python3/vimspector/debug_session.py
+++ b/python3/vimspector/debug_session.py
@@ -435,7 +435,10 @@ class DebugSession( object ):
       # working directory.
       'cwd': os.getcwd,
       'unusedLocalPort': utils.GetUnusedLocalPort,
-      'selectProcess': _SelectProcess,
+
+      # The following, starting with uppercase letters, are 'functions' taking
+      # arguments.
+      'SelectProcess': _SelectProcess,
     }
 
     # Pretend that vars passed to the launch command were typed in by the user
@@ -1597,6 +1600,8 @@ class DebugSession( object ):
       if attach_config[ 'pidSelect' ] == 'ask':
         prop = attach_config[ 'pidProperty' ]
         if prop not in launch_config:
+          # NOTE: We require that any custom picker process handles no-arguments
+          # as well as any arguments supplied in the config.
           pid = _SelectProcess()
           if pid is None:
             return
@@ -2207,8 +2212,16 @@ def PathsToAllConfigFiles( vimspector_base, current_file, filetypes ):
                                 os.path.dirname( current_file ) )
 
 
-def _SelectProcess():
-  custom_picker = settings.Get( 'custom_process_picker_expr' )
+def _SelectProcess( *args ):
+  custom_picker = settings.Get( 'custom_process_picker_func' )
   if custom_picker:
-    return int( vim.eval( custom_picker ) )
-  return int( utils.AskForInput( 'Enter Process ID: ' ) )
+    try:
+      return int( utils.Call( custom_picker, *args ) )
+    except vim.error:
+      pass
+
+  value = utils.AskForInput( 'Enter Process ID: ' )
+  if value is not None:
+    return int( value )
+
+  return value
--- a/python3/vimspector/settings.py
+++ b/python3/vimspector/settings.py
@@ -109,7 +109,7 @@ DEFAULTS = {
 
   # Custom
   'java_hotcodereplace_mode': 'ask',
-  'custom_process_picker_expr': '',
+  'custom_process_picker_func': '',
 
   # Debug configs
   'adapters': {},
--- a/python3/vimspector/settings.py
+++ b/python3/vimspector/settings.py
@@ -109,6 +109,7 @@ DEFAULTS = {
 
   # Custom
   'java_hotcodereplace_mode': 'ask',
+  'custom_process_picker_expr': '',
 
   # Debug configs
   'adapters': {},
--- a/python3/vimspector/utils.py
+++ b/python3/vimspector/utils.py
@@ -583,7 +583,9 @@ VAR_MATCH = re.compile(
     \$(?:                               # A dollar, followed by...
       (?P<escaped>\$)                |  # Another dollar = escaped
       (?P<named>[_a-z][_a-z0-9]*)    |  # or An identifier - named param
-      {(?P<braced>[_a-z][_a-z0-9]*)} |  # or An {identifier} - braced param
+                # or An {identifier} - braced param
+                # or An {identifier([arg,...])} - braced calculus call
+      {(?P<braced>[_a-z][_a-z0-9]*)(?:\((?P<args>[^\)]*)\))?} |
       {(?P<braceddefault>               # or An {id:default} - default param, as
         (?P<defname>[_a-z][_a-z0-9]*)   #   an ID
         :                               #   then a colon
@@ -596,19 +598,42 @@ VAR_MATCH = re.compile(
 
 
 class MissingSubstitution( Exception ):
-  def __init__( self, name, default_value = None ):
+  def __init__( self, name, default_value = None, args = (), arg_hash = '' ):
     self.name = name
     self.default_value = default_value
+    self.args = args
+    self.arg_hash = arg_hash
 
 
 def _Substitute( template, mapping ):
   def convert( mo ):
     # Check the most common path first.
+    args: str = mo.group( 'args' )
+    if args is None:
+      args = ()
+      arg_hash = ''
+    else:
+      # The args string must be the "internal" json of a list
+      # The key in mapping is just the calculus function a hash of the string
+      # representation of the arguments. This is important because things like
+      # unusedLocalPort should always return the same value. As a consequence
+      # multiple "calls" to the same calculus function with the same arguments
+      # only trigger one actual call.. shrug?
+      arg_hash = str( hash( args ) )
+      args = ( arg for arg in json.loads( "[" + args + "]" ) )
+
     named = mo.group( 'named' ) or mo.group( 'braced' )
     if named is not None:
-      if named not in mapping:
-        raise MissingSubstitution( named )
-      return str( mapping[ named ] )
+      if named + arg_hash not in mapping:
+        raise MissingSubstitution( named,
+                                   args = args,
+                                   arg_hash = arg_hash )
+
+      _logger.debug( "Returning %s from the map for %s with args %s",
+                     named + arg_hash,
+                     named,
+                     args )
+      return str( mapping[ named + arg_hash ] )
 
     if mo.group( 'escaped' ) is not None:
       return '$'
@@ -618,7 +643,7 @@ def _Substitute( template, mapping ):
       if named not in mapping:
         raise MissingSubstitution(
           named,
-          mo.group( 'default' ).replace( '\\}', '}' ) )
+          default_value = mo.group( 'default' ).replace( '\\}', '}' ) )
       return str( mapping[ named ] )
 
     if mo.group( 'invalid' ) is not None:
@@ -638,19 +663,25 @@ def ExpandReferencesInString( orig_s,
 
   # Parse any variables passed in in mapping, and ask for any that weren't,
   # storing the result in mapping
-  bug_catcher = 0
-  while bug_catcher < 100:
-    ++bug_catcher
-
+  while True:
     try:
       s = _Substitute( s, mapping )
       break
     except MissingSubstitution as e:
-      key = e.name
+      key = e.name + e.arg_hash
 
-      if key in calculus:
-        mapping[ key ] = calculus[ key ]()
+      if e.name in calculus:
+        mapping[ key ] = calculus[ e.name ]( *e.args )
+        _logger.debug( "Put %s into mapping for %s with args %s",
+                       key,
+                       e.name,
+                       e.args )
+      elif e.args:
+        raise ValueError( f"Invalid arguments '{ e.args }' supplied for named "
+                          f"variable '{ e.name }'. This varibale does not take "
+                          "formal arguments" )
       else:
+        assert key == e.name
         default_value = user_choices.get( key )
         # Allow _one_ level of additional substitution. This allows a very real
         # use case of "program": ${program:${file\\}}
--- a/support/custom_process_picker_fzf.vim
+++ b/support/custom_process_picker_fzf.vim
@@ -1,0 +1,22 @@
+scriptencoding utf-8
+
+let cpo = &cpo
+set cpo&
+
+" Custom process picker {{{
+
+function! PickProcessWithFzf( ... ) abort
+  if a:0 == 0
+    let source = 'ps -e'
+  else
+    let source = 'ps -e | grep ' .. a:1
+  endif
+
+  return split( fzf#run( fzf#wrap( { 'source': source }, 1 ) )[ 0 ] )[ 0 ]
+endfunction
+
+let g:vimspector_custom_process_picker_func = 'PickProcessWithFzf'
+
+" }}}
+
+let &cpo=cpo
--- a/support/test/cpp/simple_c_program/.vimspector.json
+++ b/support/test/cpp/simple_c_program/.vimspector.json
@@ -87,6 +87,12 @@
         "stopAtEntry": true,
         "MIMode": "gdb"
       }
+    },
+    "cpptools-attach": {
+      "extends": "cpptools",
+      "configuration": {
+        "request": "attach"
+      }
     }
   }
 }
--- a/tests/pidpickerinjection.test.vim
+++ b/tests/pidpickerinjection.test.vim
@@ -1,0 +1,27 @@
+function! SetUp()
+  call vimspector#test#setup#SetUpWithMappings( v:null )
+endfunction
+
+function! TearDown()
+  call vimspector#test#setup#TearDown()
+endfunction
+
+function! Test_PID_Picker()
+  let fn='test_c.cpp'
+  lcd ../support/test/cpp/simple_c_program
+  exe 'edit ' . fn
+
+  function! PickProcess() abort
+    let s:process_picker_called = v:true
+    return 12345 " doesn't matter
+  endfunction
+  let g:vimspector_custom_process_picker_expr = 'PickProcess()'
+
+  let s:process_picker_called = v:false
+  call vimspector#LaunchWithSettings( { 'configuration': 'cpptools-attach' } )
+  call assert_true(s:process_picker_called)
+  call vimspector#test#setup#Reset()
+
+  lcd -
+  %bwipeout!
+endfunction
--- a/tests/pidpickerinjection.test.vim
+++ b/tests/pidpickerinjection.test.vim
@@ -6,7 +6,11 @@ function! TearDown()
   call vimspector#test#setup#TearDown()
 endfunction
 
-function! Test_PID_Picker()
+function! Test_PID_Picker_NotCalled_NoArguments_BuiltInPidSelect()
+  " Mock AskForInput and do same as below to prove the case where it's not set
+endfunction
+
+function! Test_PID_Picker_Called_NoArguments_BuiltInPidSelect_RetInt()
   let fn='test_c.cpp'
   lcd ../support/test/cpp/simple_c_program
   exe 'edit ' . fn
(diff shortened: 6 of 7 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/664)

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/664)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @Aster89 and @puremourning)

*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/664#-NAuXBIi3_DuTzYybNxe:-NAuXBIj2Zf3g1bLW542:bvuyymo) (no related file):*
We will need tests for this. 

Documentation needs updating.  

___

Aster89:
I'm totally fine with updating the doc, but I'd do it only after having the tests in place, as without those I don't think you're gonna accept anything. And... I guess I'd need at least some guidance to make the tests :D

puremourning:
there's info on the test framework here https://github.com/puremourning/vimspector/blob/master/CONTRIBUTING.md#pr-guidelines

I suspect that the way to test this is to write a test that sets the option to some function implemented by the test script and assert that is called etc.

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/664)* status: 0 of 4 files reviewed, 6 unresolved discussions (waiting on @Aster89 and @puremourning)

*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBktU40hxqJLYT1C4H:-NBBktU40hxqJLYT1C4I:ba14xjf) (no related file):*
Still missing documentation.

___
*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBkuqyEy9pG-4SvhDT:-NBBkuqzAxLklNZi-mQG:b-ky2x8x) (no related file):*
My preference for the test would be for it to actually run to completion and attach to something. I realise this is kind of tricky and we don't actually have any tests which do this for c/cpp today. It would require a new test binary that does something like block waiting for stdin or a socket or something, attach to it, and then modify its state so it exits (or send it a message via stdin/socket).

Like I said that's quite a bit of plumbing, so I'm easy if you don't want to.

___
___
*[`support/test/cpp/simple_c_program/.vimspector.json` line 64 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBkIRR67OPYVPtqXca:-NBBkIRR67OPYVPtqXcb:bkgwzzz) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/support/test/cpp/simple_c_program/.vimspector.json#L64)):*
> ```JSON
>       "extends": "cpptools",
>       "configuration": {
>         "request": "attach"
> ```

seems to be missing the 'program' argument and various others. 
Would much rather add a configuration that actually works.
___
*[`tests/pidpickerinjection.test.vim` line 3 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBkPK6BAaGrGBLD77n:-NBBkPK6BAaGrGBLD77o:bwzi1gw) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/tests/pidpickerinjection.test.vim#L3)):*
> ```vim
> function! SetUp()
>   call vimspector#test#setup#SetUpWithMappings( v:null )
>   function! PickProcess() abort
> ```

why declare this in SetUp ?
___
*[`tests/pidpickerinjection.test.vim` line 7 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBkZ8g3rLaCT2O1TyC:-NBBkZ8g3rLaCT2O1TyD:bu9t6pm) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/tests/pidpickerinjection.test.vim#L7)):*
> ```vim
>     return 12345 " doesn't matter
>   endfunction
>   let g:vimspector_custom_process_picker = 'PickProcess()'
> ```

prefer to set this closer to where it is tested.
___
*[`tests/pidpickerinjection.test.vim` line 15 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBl-Wd4gjKzD5Vqyqn:-NBBl-Wd4gjKzD5Vqyqo:b58yiq2) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/tests/pidpickerinjection.test.vim#L15)):*
> ```vim
> 
> function! Test_PID_Picker()
>   let fn='test_c.cpp'
> ```

I'm not sure you can use this is in the tests. these files are not built into binary, so you would need to use the ones from tests/testdata/ which are compiled by the run script.
___
*[`tests/pidpickerinjection.test.vim` line 21 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBkkzF9lFZpQIxO5oZ:-NBBkkzF9lFZpQIxO5o_:bprnfn9) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/tests/pidpickerinjection.test.vim#L21)):*
> ```vim
>   let s:process_picker_called = v:false
>   call vimspector#LaunchWithSettings( { 'configuration': 'cpptools-attach' } )
>   call assert_equal(s:process_picker_called, v:true)
> ```

assert_true maybe?

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/664)* status: 0 of 4 files reviewed, 7 unresolved discussions (waiting on @Aster89 and @puremourning)

___
*[`python3/vimspector/debug_session.py` line 1324 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBle6s-8svrGhe9Y1O:-NBBle6s-8svrGhe9Y1P:b-f6pqpr) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/python3/vimspector/debug_session.py#L1324)):*
> ```Python
>         prop = attach_config[ 'pidProperty' ]
>         if prop not in launch_config:
>           pid = _SelectProcess()
> ```

I wonder if there are more data we should pass in here. It would be tempting to send the entire config object so that the script can do things like look at the configuration to say default to finding a process with a cmdline matching "program". That might be tricky if we haven't yet parsed the full config.

Aster89 (commented):
This UI is not extremely friendly.

> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/664#-:-NBDZVG18lJeslXUxQ0V:bicbeg8)* status: 0 of 4 files reviewed, 7 unresolved discussions (waiting on @puremourning)

*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBktU40hxqJLYT1C4H:-NBDU99E2MXFzU81ZdHL:bh9uxr5) (no related file):*

Still missing documentation.
</blockquote></details>

I promise I'll write it.

___
*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBkuqyEy9pG-4SvhDT:-NBDUXRP0iAS_Cvo_Gk4:b-cuct5f) (no related file):*

My preference for the test would be for it to actually run to completion and attach to something. I realise this is kind of tricky and we don't actually have any tests which do this for c/cpp today. It would require a new test binary that does something like block waiting for stdin or a socket or something, attach to it, and then modify its state so it exits (or send it a message via stdin/socket).

Like I said that's quite a bit of plumbing, so I'm easy if you don't want to.
</blockquote></details>

It does look quite a bit of work indeed, but... why doing it? I mean, from the moment `_SelectProcess()` returns an `int` (and that's what it returns if it returns at all, right?), all the rest of the code is tested. In the sense that all the existing code should be reliant to one giving a wrong PID or not even a number. As in, why should we have a test that actually attaches to something, if the user is free to define a function which returns the wrong number or not even a number, which would result in a failure to attach? It's a bit like providing a slingshot as a throw-somthing tool, and then using stones with the label "for-sligshot" on them, whereas you could just use a coin, a snotball or whatever you have at hand, no?

___
___
*[`python3/vimspector/debug_session.py` line 1324 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBle6s-8svrGhe9Y1O:-NBDXwfb3qksFTbNOoTi:b-ugapaw) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/python3/vimspector/debug_session.py#L1324)):*

I wonder if there are more data we should pass in here. It would be tempting to send the entire config object so that the script can do things like look at the configuration to say default to finding a process with a cmdline matching "program". That might be tricky if we haven't yet parsed the full config.
</blockquote></details>

If I've understood what you mean, that looks like an interesting idea. But wouldn't it be an extension to the behavior provided with the current change? As in, couldn't I do it with another PR in the future?
___
*[`support/test/cpp/simple_c_program/.vimspector.json` line 64 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBkIRR67OPYVPtqXca:-NBDYFGR46j4y_x8LTV2:bl6scz4) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/support/test/cpp/simple_c_program/.vimspector.json#L64)):*

seems to be missing the 'program' argument and various others. 
Would much rather add a configuration that actually works.
</blockquote></details>

Am I not extending `cpptools`? Don't I inherit the other part of the JSON snippet?
___
*[`tests/pidpickerinjection.test.vim` line 3 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBkPK6BAaGrGBLD77n:-NBDYd8V9DdkK16OGpAa:b45h) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/tests/pidpickerinjection.test.vim#L3)):*

why declare this in SetUp ?
</blockquote></details>





_Suggestion:_
```vim
As regards `call vimspector#test#setup#SetUpWithMappings( v:null )`, maybe I don't need. Maybe it was in the stuff I copied from :D

As regards `function! PickProcess() abort`, do you mean that the whole funciton should be defined elsewhere? If so, where?
```
___
*[`tests/pidpickerinjection.test.vim` line 7 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBkZ8g3rLaCT2O1TyC:-NBDYscrAACAqbFBnYiE:b-3f8t7a) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/tests/pidpickerinjection.test.vim#L7)):*

prefer to set this closer to where it is tested.
</blockquote></details>

Does this answer my question above?
___
*[`tests/pidpickerinjection.test.vim` line 15 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBl-Wd4gjKzD5Vqyqn:-NBDYzMy4GfpMndCA8Ay:bij7b12) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/tests/pidpickerinjection.test.vim#L15)):*

I'm not sure you can use this is in the tests. these files are not built into binary, so you would need to use the ones from tests/testdata/ which are compiled by the run script.
</blockquote></details>

See my longest comment. I don't understand why we need a binary to debug at all.

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/664)* status: 0 of 4 files reviewed, 6 unresolved discussions (waiting on @Aster89 and @puremourning)

___
*[`tests/pidpickerinjection.test.vim` line 15 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBl-Wd4gjKzD5Vqyqn:-NBNjb2eEJLFgiQYt8et:bft4dyy) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/tests/pidpickerinjection.test.vim#L15)):*
<details><summary><i>Previously, Aster89 (Enrico Maria De Angelis) wrote…</i></summary><blockquote>

See my longest comment. I don't understand why we need a binary to debug at all.
</blockquote></details>

all this tests is that the function is called, not that the functionality _works_

Aster89 (commented):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/664)* status: 0 of 4 files reviewed, 6 unresolved discussions (waiting on @puremourning)

___
*[`python3/vimspector/debug_session.py` line 1324 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBle6s-8svrGhe9Y1O:-NBNrXL7-PqhVCLYmI05:bdje47k) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/python3/vimspector/debug_session.py#L1324)):*

changing the API in future would be a breaking change, so we would have to account for at least the possibility of passing arguments now.
</blockquote></details>

Oh, I see. I'll look into it. Thanks for the explanation.
___
*[`tests/pidpickerinjection.test.vim` line 15 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBl-Wd4gjKzD5Vqyqn:-NBNs2Gv15_lq8-GpLZ6:b-vxkdjv) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/tests/pidpickerinjection.test.vim#L15)):*

all this tests is that the function is called, not that the functionality _works_
</blockquote></details>

Well, if the function is to be defined by the user, what can we possibly test about it? All we care is that we call it, what it does... it's none of our business, no?
___
*[`tests/pidpickerinjection.test.vim` line 21 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBkkzF9lFZpQIxO5oZ:-NBNrSBL-eAIxfWsroVe:bb7f2mf) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/tests/pidpickerinjection.test.vim#L21)):*

assert_true maybe?
</blockquote></details>

done

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/664)* status: 0 of 4 files reviewed, 6 unresolved discussions (waiting on @Aster89 and @puremourning)

___
*[`support/test/cpp/simple_c_program/.vimspector.json` line 64 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBkIRR67OPYVPtqXca:-NBNsaS2CPxxT7MltAV7:b9kr5n0) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/support/test/cpp/simple_c_program/.vimspector.json#L64)):*
<details><summary><i>Previously, Aster89 (Enrico Maria De Angelis) wrote…</i></summary><blockquote>

Am I not extending `cpptools`? Don't I inherit the other part of the JSON snippet?
</blockquote></details>

oh yeah, sorry , missed that

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/664)* status: 0 of 4 files reviewed, 5 unresolved discussions (waiting on @Aster89 and @puremourning)

___
*[`tests/pidpickerinjection.test.vim` line 7 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBkZ8g3rLaCT2O1TyC:-NBNsgvK6vt9TLzhKSGJ:b-alznt0) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/tests/pidpickerinjection.test.vim#L7)):*
<details><summary><i>Previously, Aster89 (Enrico Maria De Angelis) wrote…</i></summary><blockquote>

Does this answer my question above?
</blockquote></details>

move the definition and setting into the test function 
___
*[`tests/pidpickerinjection.test.vim` line 15 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBl-Wd4gjKzD5Vqyqn:-NBNsliRABp96rOb--7p:bd5nwp3) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/tests/pidpickerinjection.test.vim#L15)):*
<details><summary><i>Previously, Aster89 (Enrico Maria De Angelis) wrote…</i></summary><blockquote>

Well, if the function is to be defined by the user, what can we possibly test about it? All we care is that we call it, what it does... it's none of our business, no?
</blockquote></details>

well I was hoping that we could test that we actually use the result.

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/664)* status: 0 of 4 files reviewed, 6 unresolved discussions (waiting on @Aster89 and @puremourning)

___
*[`python3/vimspector/debug_session.py` line 1881 at r3](https://reviewable.io/reviews/puremourning/vimspector/664#-NBOQZV2A03pQ0Aifno7:-NBOQZV2A03pQ0Aifno8:bh4u236) ([raw file](https://github.com/puremourning/vimspector/blob/249c7506aaf3a1d31ffae21126402c86feb99b02/python3/vimspector/debug_session.py#L1881)):*
> ```Python
> 
> def _SelectProcess():
>   custom_picker = settings.Get( 'custom_process_picker' )
> ```

I just tried this out. it turns out that custom_process_picker is an _expression_ (rather than strictly a function), so we might want to call the option 'custom_process_picker_expr' to make ti clear that it's an expression (and therefore requires `()` for function calls)

otherwise the change works well. I even used your fzf code, which worked, but made me really want it to "pre filter" for the app I'm testing, so like `${selectProcess( 'ycmd' )}` - this is potentially a lot of work though

Aster89 (commented):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/664)* status: 0 of 4 files reviewed, 6 unresolved discussions (waiting on @puremourning)

___
*[`python3/vimspector/debug_session.py` line 1881 at r3](https://reviewable.io/reviews/puremourning/vimspector/664#-NBOQZV2A03pQ0Aifno7:-NBOjhQsAe7zri69dJNb:bmsgip0) ([raw file](https://github.com/puremourning/vimspector/blob/249c7506aaf3a1d31ffae21126402c86feb99b02/python3/vimspector/debug_session.py#L1881)):*

I just tried this out. it turns out that custom_process_picker is an _expression_ (rather than strictly a function), so we might want to call the option 'custom_process_picker_expr' to make ti clear that it's an expression (and therefore requires `()` for function calls)

otherwise the change works well. I even used your fzf code, which worked, but made me really want it to "pre filter" for the app I'm testing, so like `${selectProcess( 'ycmd' )}` - this is potentially a lot of work though
</blockquote></details>

renamed.

As regards the other points, would `${selectProcess(  'ycmd' )}` something you write where? In the `.vimspector.json`?
___
*[`tests/pidpickerinjection.test.vim` line 7 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBkZ8g3rLaCT2O1TyC:-NBOjDmW7GfCWmMU1lR3:b-896fix) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/tests/pidpickerinjection.test.vim#L7)):*

move the definition and setting into the test function 
</blockquote></details>

Done.
___
*[`tests/pidpickerinjection.test.vim` line 15 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBl-Wd4gjKzD5Vqyqn:-NBOkWMI34qs7bnd1M3i:bj6tikr) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/tests/pidpickerinjection.test.vim#L15)):*

well I was hoping that we could test that we actually use the result.
</blockquote></details>

Result of what? `_SelectProcess()`? If so, then the usage is at the line `pid = _SelectProcess()`, and that was `pid = utils.AskForInput( 'Enter PID to attach to: ' )`.  So if you have code in place testing that the result of `utils.AskForInput( 'Enter PID to attach to: ' )` was used, that code is testing that the result of `_SelectProcess()` is used, no?

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/664)* status: 0 of 4 files reviewed, 6 unresolved discussions (waiting on @Aster89 and @puremourning)

___
*[`tests/pidpickerinjection.test.vim` line 15 at r2](https://reviewable.io/reviews/puremourning/vimspector/664#-NBBl-Wd4gjKzD5Vqyqn:-NBTTkFK0imQrGaNGp05:b-9urgqk) ([raw file](https://github.com/puremourning/vimspector/blob/f3481c4855d4412691337598dc23156d033f1352/tests/pidpickerinjection.test.vim#L15)):*
<details><summary><i>Previously, Aster89 (Enrico Maria De Angelis) wrote…</i></summary><blockquote>

Result of what? `_SelectProcess()`? If so, then the usage is at the line `pid = _SelectProcess()`, and that was `pid = utils.AskForInput( 'Enter PID to attach to: ' )`.  So if you have code in place testing that the result of `utils.AskForInput( 'Enter PID to attach to: ' )` was used, that code is testing that the result of `_SelectProcess()` is used, no?
</blockquote></details>

you're making (in my opinion) a fundamental testing error - you're testing the code, not the product.

To put it a bit more bluntly, I expect the tests to be more like "integration" tests than "module" or "white box" tests. I only care that the functionality _works_ from the user perspective, because that's what I have to support.

For example, a set of meaningful tests might be:

When I have a custom PID picker, I can pick supply a PID.

When I supply a custom PID, and:

1. pidselect = ask, vimspector inserts that into the configuration sent to the debug adapter (ideally: and the debug adapter successfully attaches to the process)
2. the configuration contains "${selectProcess}", vimspector inserts that into the configuration sent to the debug adapter 
3. the configuration contains "${selectProcess:some default value}", vimspector ? what? 
4. the configuration contains "${selectProcess}" coerced to some type, vimspector inserts that into the configuration sent to the debug adapter
5. the custom selector returns invalid data, ?? what???
6. the custom selector raises an error, ??? what???
7. the user presses ctrl-c when the custom selector is visible, ?? what ???


This is the kind of level of testing I typically want. Writing tests that prove the code does what the code does is kind of pointless.

puremourning (commented):
> Reviewed 1 of 2 files at r2, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/664)* status: all files reviewed, 6 unresolved discussions (waiting on @Aster89 and @puremourning)

Aster89:
I've just tried it. It looks ok to me. Always ready to write the doc.

python3/vimspector/debug_session.py
```diff
@@ -1975,8 +1980,16 @@ def PathsToAllConfigFiles( vimspector_base, current_file, filetypes ):
...
-  custom_picker = settings.Get( 'custom_process_picker_expr' )
+def _SelectProcess( *args ):
+  custom_picker = settings.Get( 'custom_process_picker_func' )
   if custom_picker:
-    return int( vim.eval( custom_picker ) )
-  return int( utils.AskForInput( 'Enter Process ID: ' ) )
+    try:
+      return int( utils.Call( custom_picker, *args ) )
+    except vim.error:
+      pass
```
Aster89, line 1989: Why? If I understand correctly, this falls back on the normal "enter the pid by hand" behavior, but why should we do it? If the pid pickers is broken, it is broken. The user goes to its vimrc and fixes it, no?

python3/vimspector/debug_session.py
```diff
@@ -1410,6 +1413,8 @@ def _PrepareAttach( self, adapter_config, launch_config ):
       if attach_config[ 'pidSelect' ] == 'ask':
         prop = attach_config[ 'pidProperty' ]
         if prop not in launch_config:
+          # NOTE: We require that any custom picker process handles no-arguments
+          # as well as any arguments supplied in the config.
           pid = _SelectProcess()
```
Aster89, line 1418: Is this meant to be passed something (maybe on a future commit in before accepting the PR)? If not, then why does the definition take `*args`?

Aster89 (commented):
Sorry for the possibly dumb questions, but I've lost track of why the change looks like it looks now. I see it's mainly because of tests, no? Anyway, I'm always available to write the doc.

puremourning:
Just fyi I haven't forgotten about this, I still intend to so _something_ here, I'm just still not sure what is best.

Aster89:
Actually hold on, because I had another thought about it. I'll find some time to write to you in the chat in the afternoon.

PR #668, opened 2022-09-09
Fix type error in configuration example

danielrainer:
The value of `configurations.<config-name>.configuration.program` must have type string and should only be the executable. Arguments are passed via the value of `configurations.<config-name>.configuration.args`.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -136,8 +136,8 @@ But for now, consider the following example snippet:
       },
       "configuration": {
         "request": "launch" /* or it could be "attach" */,
-        "program": [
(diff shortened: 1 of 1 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/668)

puremourning:
Thanks for sending a PR, but honestly, I don't quite get it. You're correcting and example/completely fictitious configuration?

danielrainer:
It seems that I misunderstood these configuration option. I thought that they are adapter-independent but now that I looked into it I see that these are adapter-specific. Thank you for the quick clarification.

PR #683, opened 2022-10-12
Instruction breakpoints

puremourning:
Add support for instruction breakpoints to the disassembly window.

Bit flaky, but work reasonably as a quick win.

We model them as line breakpoints where the file/line point to the disassembly buffer. Then we use the address of the instruction associated with the breakpoint to find the new line whenever the disassembly changes.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -46,6 +46,9 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
+    - name: 'Install requirements'
+      run: pip3 install --user -r dev_requirements.txt
+
     - run: |
         go install github.com/go-delve/delve/cmd/dlv@latest
       name: 'Install Delve for Go'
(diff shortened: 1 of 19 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/683)

puremourning:
https://asciinema.org/a/esEncAxP45CJmo8Em1sQtxRYe

demo

puremourning:
VScode clearly uses the address as the key, and _does_ persist them across restarts of the process. This may lead to unpredictable results. But I need to check what gdb and lldb do from the command line, as that's what most people would expect.

We could also ship this without persistence and refactor later, as it should be a fairly simple obviously-better or not. 

Need to test save/restore too - what do we sae there.

puremourning (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/683#-:-NF2t8qu7aCyXMU15iCl:bnfp4nl)* status: 0 of 19 files reviewed, all discussions resolved

puremourning (commented):
> Reviewed 1 of 5 files at r1, 18 of 18 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/683)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @puremourning)

PR #696, opened 2022-10-28
Disassembly scoll

puremourning:
"Infinite" scrolling for the disassembly window.

The way this works is it dumbly requests an extra page when the top line of the window reaches 1, and when its >= buffer length minus 1 page.

We don't trim or anything, so it just grows. It always shrinks the next time we jump to a frame though.

We also drop any request if one is in progress, which is not ideal, but it's simple.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1218,8 +1218,11 @@ use the default "step" commands (e.g. `<F10>`), the stepping is automatically
 chnged to per-instruction rather than per statement.
 
 Each time the process stops, vimspector requests about 2 windows full of
-instructions around the current PC. To see more, you have to make the window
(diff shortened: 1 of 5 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/696)

puremourning (commented):
:lgtm:

> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/696#-:-NFURY7_DP6P4mi4EBdK:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @puremourning)

PR #722, opened 2023-02-19
feat: add timeout

zhoupro:
When debug delve, It will show  lauch failed msg, then debug begin. I looked into it, find no timeout param.

### DIFF
--- a/python3/vimspector/debug_session.py
+++ b/python3/vimspector/debug_session.py
@@ -1623,8 +1623,7 @@ class DebugSession( object ):
         'arguments': self._launch_config
       },
       failure_handler,
-      timeout = self._connection.sync_timeout
-      )
+      timeout = self._connection.sync_timeout )
 
 
   def _OnLaunchComplete( self ):
--- a/python3/vimspector/debug_session.py
+++ b/python3/vimspector/debug_session.py
@@ -1622,7 +1622,9 @@ class DebugSession( object ):
         'command': self._launch_config[ 'request' ],
         'arguments': self._launch_config
       },
-      failure_handler )
+      failure_handler,
+      timeout = self._connection.sync_timeout
+      )
 
 
   def _OnLaunchComplete( self ):


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/722)

puremourning (changes requested):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/722)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @zhoupro)

___
*[`python3/vimspector/debug_session.py` line 1626 at r2](https://reviewable.io/reviews/puremourning/vimspector/722#-NQG99SHAdVoElTb1sXn:-NQG99SHAdVoElTb1sXo:b-r5xtjo) ([raw file](https://github.com/puremourning/vimspector/blob/6d2233ec48365fdccf5946c3aa98712678c5f50e/python3/vimspector/debug_session.py#L1626)):*
> ```Python
>       },
>       failure_handler,
>       timeout = self._connection.sync_timeout )
> ```

If no timeout is supplied here, it would use the `async_timeout` , which is by default 15 seconds. By setting this to `sync_timeout` it will reduce that to 5 seconds.

so I'm confused why this change solves the purported problem?

zhoupro:
I say my case  first.
I use flollowing config, which includ `sync_timeout` and `iasync_timeout` and also include a `timeout`.  Timeouts are founded in issues.    But when debug a very slow program, timeout msg occured. However after a while. the real debug session come.  So i find file based message, found no specifc time. So I just add timeout which can be found in same file.  after that, The problem is solved.  
`{
  "configurations": {
      "test": {
      "adapter": {
        "extends":"delve",
        "sync_timeout":1500000,
        "iasync_timeout":1500000
      },
      "configuration": {
        .....
        "timeout": 3000000,
        .....
    }
  }
}`

secondly. I alse specfic `sync_timeout`, which is greater than 15s. after that my probem solved.  @puremourning

puremourning:
> iasync_timeout

seems incorrect. I'm still not sure about this PR and without a solid repro test and explanation for why it's actually needed, then I think we can close it.

PR #726, opened 2023-03-02
Update README.md (fix bash name)

Constantin1489:
Bash is an acronym for Bourne Again Shell. Reference: https://www.gnu.org/gnu/thegnuproject.html


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/726)

puremourning:
Thanks for sending a PR! While strictly true, I don't think this change significantly improves the clarity or understanding for users, so I'm inclined to leave it as it is.

PR #732, opened 2023-03-19
Add `CollapseContainer` function to collapse all variables within a container

Danthewaann:
Hi, thanks for the great plugin!

I've made a small addition to vimspector to support collapsing a variable container in the variables window. (It also works in the watches window and I think the disassemble window. I haven't tested the disassemble window as I use Python which doesn't support disassembly. I can add tests for these if requested)

My problem is that when I drilled down into some variables, I found it awkward to collapse the container by needing to go to the container variable and pressing enter to collapse it. I would need to do this multiple times to get back to where I wanted to be.

The change I've made is adding a new `CollapseContainer` function that executes when you press `x`. It recursively collapses all variables inside the current container where the cursor is and places the cursor at the line where the container variable is (it is recursive here as each expanded variable might contain variables that are also expanded).

I've both tested this manually with my own vim setup, and I've added a single `Test_CollapseContainer` test in `variables.test.vim` to demonstrate the change.

I'm happy for feedback and any requested changes!

Note: I did also add an extra `<Tab>` bind to `expand_collapse`, which I find easier to reach than `<CR>`. I can remove this if it is out of scope for this PR.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/732)

puremourning:
I'm gonna try this out now

puremourning (changes requested):
> Reviewed 5 of 6 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/732)* status: all files reviewed, 7 unresolved discussions (waiting on @dcrblack)

___
*[`python3/vimspector/settings.py` line 64 at r3](https://reviewable.io/reviews/puremourning/vimspector/732#-NRU3yrx1wSCzlC0ykPK:-NRU3yrx1wSCzlC0ykPL:besrfiv) ([raw file](https://github.com/puremourning/vimspector/blob/b86d5bbc985100e551dfe44c21842ffe6bb22e0b/python3/vimspector/settings.py#L64)):*
> ```Python
>   'mappings': {
>     'variables': {
>       'expand_collapse': [ '<CR>', '<2-LeftMouse>', '<Tab>' ],
> ```

needs documentation update
___
*[`python3/vimspector/settings.py` line 65 at r3](https://reviewable.io/reviews/puremourning/vimspector/732#-NRU3-bV5x8rilsRm09Z:-NRU3-bV5x8rilsRm09_:b3xiw7d) ([raw file](https://github.com/puremourning/vimspector/blob/b86d5bbc985100e551dfe44c21842ffe6bb22e0b/python3/vimspector/settings.py#L65)):*
> ```Python
>     'variables': {
>       'expand_collapse': [ '<CR>', '<2-LeftMouse>', '<Tab>' ],
>       'collapse': [ 'x' ],
> ```

'x' feels like "delete" to me. I'm not sure that's the ideal default mapping.

do you know of any other UI "trees" that have this shortcut? If so, what do they use? (I mean in any UI paradigm). My initial inclination was something like shift-tab. 

Then I thought about vim paradigms and the best parallel I could think of was folding. for folding you have `z0` to "close al folds under the cursor recursively", though I personally use `zA` which is more of a toggle. I think trying to map all of the numerous `z*` commands is far fetched, but I _do_ think it is worth thinking hard about what the default should be and having some logic or intuition behind it.


Also, need documentation update.
___
*[`python3/vimspector/settings.py` line 65 at r3](https://reviewable.io/reviews/puremourning/vimspector/732#-NRU43kE2cMdnOBXoy5R:-NRU43kE2cMdnOBXoy5S:b-ojvpyy) ([raw file](https://github.com/puremourning/vimspector/blob/b86d5bbc985100e551dfe44c21842ffe6bb22e0b/python3/vimspector/settings.py#L65)):*
> ```Python
>     'variables': {
>       'expand_collapse': [ '<CR>', '<2-LeftMouse>', '<Tab>' ],
>       'collapse': [ 'x' ],
> ```

maybe call it "collapse_container" or something so as not to confuse the expand_collapse above?
___
*[`python3/vimspector/variables.py` line 604 at r3](https://reviewable.io/reviews/puremourning/vimspector/732#-NRU5aAL2D4xKSvhtCwX:-NRU5aAL2D4xKSvhtCwY:bb2sk7e) ([raw file](https://github.com/puremourning/vimspector/blob/b86d5bbc985100e551dfe44c21842ffe6bb22e0b/python3/vimspector/variables.py#L604)):*
> ```Python
>         # Make sure this variable and all it's variables are
>         # collapsed before continuing
>         view.draw()
> ```

this is _extremely_ inefficient. the draw call is very expensive - it clears and re-draws the buffer.

It's not clear why this is needed. The comment states the _what_ not the _why_.
___
*[`python3/vimspector/variables.py` line 622 at r3](https://reviewable.io/reviews/puremourning/vimspector/732#-NRU5-mPB5HRgiPBhMDM:-NRU5-mPB5HRgiPBhMDN:b-wnoh4o) ([raw file](https://github.com/puremourning/vimspector/blob/b86d5bbc985100e551dfe44c21842ffe6bb22e0b/python3/vimspector/variables.py#L622)):*
> ```Python
>       self._CollapseVariables( container.variables, view )
> 
>     for i in range( line_num, 0, -1 ):
> ```

line_num should never be 0, it's 1-based

I'm sort of struggling to work out what the purpose of this loop is. In the previous stanza we collapsed all the variables below the "parent" of the current one (i.e. of `container`). Now we are scanning the buffer backwards to find `container` ? I guess I'm unsure why we need to do this. can't we just do `container.exanded = Expandable.COLLAPSED_BY_USER` ?
___
*[`python3/vimspector/variables.py` line 630 at r3](https://reviewable.io/reviews/puremourning/vimspector/732#-NRU5MMDEb1Udjtug0h-:-NRU5MMDEb1Udjtug0h0:b-tdn41v) ([raw file](https://github.com/puremourning/vimspector/blob/b86d5bbc985100e551dfe44c21842ffe6bb22e0b/python3/vimspector/variables.py#L630)):*
> ```Python
>       if id( container ) == id( variable ):
>         variable.expanded = Expandable.COLLAPSED_BY_USER
>         vim.current.window.cursor = ( i, 0 )
> ```

I'm not sure we would want to do this inside the popup, because the current window cannot be the popup

Is it even valid to set cursor column to 0 ? 
___
*[`tests/variables.test.vim` line 358 at r3](https://reviewable.io/reviews/puremourning/vimspector/732#-NRUC4UB96mjnCMjJ1GL:-NRUC4UB96mjnCMjJ1GM:bqcyt5q) ([raw file](https://github.com/puremourning/vimspector/blob/b86d5bbc985100e551dfe44c21842ffe6bb22e0b/tests/variables.test.vim#L358)):*
> ```vim
> endfunction
>  
> function! Test_CollapseContainer()
> ```

I think we need the tests to cover Watches and hovers too because there are subtle differences.

puremourning (commented):
Thanks for the PR! 

I'm in favour of the behaviour. Just a bit of polish I think.

> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/732#-:-NRUCG7c4wTxX54wHc6q:b18wtvu)* status: all files reviewed, 7 unresolved discussions (waiting on @dcrblack)

dcrblack (commented):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/732)* status: all files reviewed, 7 unresolved discussions (waiting on @puremourning)

___
*[`python3/vimspector/settings.py` line 65 at r3](https://reviewable.io/reviews/puremourning/vimspector/732#-NRU43kE2cMdnOBXoy5R:-NRgTaF_Bj3q1f7Juh0l:b-5thqby) ([raw file](https://github.com/puremourning/vimspector/blob/b86d5bbc985100e551dfe44c21842ffe6bb22e0b/python3/vimspector/settings.py#L65)):*

maybe call it "collapse_container" or something so as not to confuse the expand_collapse above?
</blockquote></details>

That makes sense to me.
___
*[`python3/vimspector/settings.py` line 65 at r3](https://reviewable.io/reviews/puremourning/vimspector/732#-NRU3-bV5x8rilsRm09Z:-NRgTd6j5dRDfiL_MA1k:bt5wnv8) ([raw file](https://github.com/puremourning/vimspector/blob/b86d5bbc985100e551dfe44c21842ffe6bb22e0b/python3/vimspector/settings.py#L65)):*

'x' feels like "delete" to me. I'm not sure that's the ideal default mapping.

do you know of any other UI "trees" that have this shortcut? If so, what do they use? (I mean in any UI paradigm). My initial inclination was something like shift-tab. 

Then I thought about vim paradigms and the best parallel I could think of was folding. for folding you have `z0` to "close al folds under the cursor recursively", though I personally use `zA` which is more of a toggle. I think trying to map all of the numerous `z*` commands is far fetched, but I _do_ think it is worth thinking hard about what the default should be and having some logic or intuition behind it.


Also, need documentation update.
</blockquote></details>

I got the idea to use `x` from NERDTree, as it's used in that plugin to close the current directory tree where the cursor is. I think `x` in this case means `close` more so than `delete`. Like the `x` button you would see in web browser tab to close it. 
___
*[`python3/vimspector/variables.py` line 604 at r3](https://reviewable.io/reviews/puremourning/vimspector/732#-NRU5aAL2D4xKSvhtCwX:-NRgRBev9hdn4UMVYZ0R:bu3tirn) ([raw file](https://github.com/puremourning/vimspector/blob/b86d5bbc985100e551dfe44c21842ffe6bb22e0b/python3/vimspector/variables.py#L604)):*

this is _extremely_ inefficient. the draw call is very expensive - it clears and re-draws the buffer.

It's not clear why this is needed. The comment states the _what_ not the _why_.
</blockquote></details>

It turns out this draw call isn't needed. I must've added it as an attempted fix for redrawing nested containers. I've removed the draw as it doesn't affect functionality
___
*[`python3/vimspector/variables.py` line 622 at r3](https://reviewable.io/reviews/puremourning/vimspector/732#-NRU5-mPB5HRgiPBhMDM:-NRgRUFx5CqTlhyryCW4:b-hlkx0w) ([raw file](https://github.com/puremourning/vimspector/blob/b86d5bbc985100e551dfe44c21842ffe6bb22e0b/python3/vimspector/variables.py#L622)):*

line_num should never be 0, it's 1-based

I'm sort of struggling to work out what the purpose of this loop is. In the previous stanza we collapsed all the variables below the "parent" of the current one (i.e. of `container`). Now we are scanning the buffer backwards to find `container` ? I guess I'm unsure why we need to do this. can't we just do `container.exanded = Expandable.COLLAPSED_BY_USER` ?
</blockquote></details>

The only purpose for this loop is to get the line number where `container` is located in the buffer so we can set the current cursor position to it. This just makes easily to collapse multiple nested containers when pressing `x` multiple times  without needing to move to cursor. I'll need to test this particular change out more, especially in the variable popup window.
___
*[`python3/vimspector/variables.py` line 630 at r3](https://reviewable.io/reviews/puremourning/vimspector/732#-NRU5MMDEb1Udjtug0h-:-NRgSIQ_C_iuXAxgYpsQ:bp16asl) ([raw file](https://github.com/puremourning/vimspector/blob/b86d5bbc985100e551dfe44c21842ffe6bb22e0b/python3/vimspector/variables.py#L630)):*

I'm not sure we would want to do this inside the popup, because the current window cannot be the popup

Is it even valid to set cursor column to 0 ? 
</blockquote></details>

I haven't explicitly tested the variable popup window. I'll look into testing my changes out with the popup window.

Issue #679, opened 2022-10-05
Folds are not opened when jumping to a frame


puremourning:
### Description

If you have folding enabled for the filetype being debugged and the current PC happens to be within a fold, the fold is not expanded when jumping to that location.

### Works in vim?

Reproducable in Vim

### Works with clean config?

Yes

### Sample project works?

Reproducable with sample project

### Minimal Reproduction

put a fold marker covering the whole test file (e.g. support/test/cpp/simple_c/test_c.cpp), e.g. `// test {{{ ... }}}`
Ensure it remains folded. (
Start debugging.

### Expected Behaviour

The fold is expanded to see the PC line

### Actual Behaviour

PC line is highlighted as the fold marker

### Installation Type

Plugin Manager (git clone)

### Vimspector version

HEAD

### Vim version

```Text
Any
```

### Declaration

- [X] I have read and understood [CONTRIBUTING.md](https://github.com/puremourning/vimspector/blob/master/CONTRIBUTING.md#diagnostics)
- [X] I have read and understood the [CODE_OF_CONDUCT.md](https://github.com/puremourning/vimspector/blob/master/CODE_OF_CONDUCT.md)

puremourning:
workaround, in VimspectorJumptedToFrame, do a `zv`

PR #745, opened 2023-04-09
doautocmd VimspectorJumpedToFrame after cursor moved

timsu92:
Before, this autocmd is called only when moved to a file but not after cursor is move as well. This PR resolves this.

This is inspired when solving https://github.com/puremourning/vimspector/issues/679#issuecomment-1268679344 and that suggestion wasn't success. That's because this wasn't executed:
https://github.com/puremourning/vimspector/blob/242764edc70808eb6ec7be5a057ca4d41c1b9329/python3/vimspector/code.py#L147-L149
However, I see this before it:
https://github.com/puremourning/vimspector/blob/242764edc70808eb6ec7be5a057ca4d41c1b9329/python3/vimspector/code.py#L140-L141
I'm not sure whether `SetCursorPosInWindow` should be located inside the `if`.

### DIFF
--- a/python3/vimspector/code.py
+++ b/python3/vimspector/code.py
@@ -132,7 +132,6 @@ class CodeView( object ):
     with utils.LetCurrentWindow( self._window ):
       try:
         utils.OpenFileInCurrentWindow( frame[ 'source' ][ 'path' ] )
-        vim.command( 'doautocmd <nomodeline> User VimspectorJumpedToFrame' )
       except vim.error:
         self._logger.exception( 'Unexpected vim error opening file {}'.format(
           frame[ 'source' ][ 'path' ] ) )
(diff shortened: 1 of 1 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/745)

timsu92:
Or, do you think I should make this line in to a try...except vim.error block？

puremourning (commented):
Thanks. Can we update tests/ui.test.vim so that it tests this new behaviour (and would fail before this change) ?

> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/745#-:-NSoqqE6A2qxy_0A6OD-:b9w0l40)* status: 0 of 1 files reviewed, all discussions resolved

PR #756, opened 2023-04-18
Support multiple independent root level sessions (tabs)

### LINKED ISSUE #749: [Feature Request] Multiple concurrent independent debug sessions (debug tabs)
Forked off the requirement for independent sessions. The current focus is on "child" sessions as required by debugpy and vscode-js-debug.

----

I'd like to have multiple debugging sessions at the same time as well for a different reason: I've got multiple Java applications in the same 'workspace' root (not sure if this is the correct word) and they communicate with each other through API requests and/or queues. Hence it'll be extremely handy if vimspector can attach to multiple applications simultaneously and set breakpoints in different sub-projects.

May I know whether my request is the same as this one or I should open a new ticket?

(Originally reported at dansomething/coc-java-debug#16)

_Originally posted by @Frederick888 in https://github.com/puremourning/vimspector/issues/141#issuecomment-745255008_

puremourning:
This is a bit clunky but works, even for 2 root level sessions which themselves have child sessions.

You have the following facilities:

1. VimspectorNewSession <name> This creates a new session and makes it active. Optional name is used in place of the generated one when starting a launch.

2. Switching to a specific debug tab makes that session active. This is intuitive and probably the most common way to work with this.

3. Switching manually using VimspectorSwitchToSession <tab complete>.

4. vimspector#GetSessionName() useful for putting in a statusline. I used this:

```
function! BenGetCustomInfo()
  if !empty( &buftype )
    return ''
  endif

  if !exists( '*vimspector#GetSessionName' )
    return ''
  endif

  return vimspector#GetSessionName()
endfunction

" ... statusline stuff
" Custom info
set statusline+=%*
set statusline+=%(\ %.20{BenGetCustomInfo()}\ %)

```

TODO: The tests will fail due to the change in name of the session buffers.

Fixes #749

### DIFF
--- a/README.md
+++ b/README.md
@@ -1060,18 +1060,22 @@ pick one.
 Vimspector can save and restore breakpoints (and some other stuff) to a session
 file. The following commands exist for that:
 
-* `VimspectorMkSession [file name]` - save the current set of line breakpoints,
+* `VimspectorMkSession [file/dir name]` - save the current set of line breakpoints,
   logpoints, conditional breakpoints, function breakpoints and exception
-  breakpoint filters to the session file.
-* `VimspectorLoadSession [file name]` - read breakpoints from the session file
-  and replace any currently set breakpoints. Prior to loading, all current
-  breakpoints are cleared (as if `vimspector#ClearLineBreakpoints()` was
-  called).
+  breakpoint filters to the supplied session file or the default file in the
+  supplied directory.
+* `VimspectorLoadSession [file/dir name]` - read breakpoints from the session
+  file supplied or the default file in the supplied directory and replace any
+  currently set breakpoints. Prior to loading, all current breakpoints are
+  cleared (as if `vimspector#ClearLineBreakpoints()` was called).
 
-In both cases, the file name argument is optional. By default, the file is named
-`.vimspector.session`, but this can be changed globally by setting
+In both cases, the file/dir name argument is optional. By default, the file is
+named `.vimspector.session`, but this can be changed globally by setting
 `g:vimspector_session_file_name` to something else, or by manually specifying a
-path when calling the command.
+path when calling the command. If you supply a directory, the default or
+configured session file name is read fron or written to that directory.
+Othewise, the file is read based on the currently open buffer or written to the
+current working directory.
 
 Advanced users may wish to automate the process of loading and saving, for
 example by adding `VimEnter` and `VimLeave` autocommands. It's recommended in
--- a/README.md
+++ b/README.md
@@ -146,6 +146,7 @@ And a couple of brief demos:
 - logging/stdout display
 - simple stable API for custom tooling (e.g. integrate with language server)
 - view hex dump of process memory
+- multiple independent debugging sessions (debug different apps in tabs)
 - multi-process (multi-session) debugging
 
 ## Supported languages
@@ -569,9 +570,15 @@ Vimspector is a vim UI on top of the Debug Adapter Protocol. It's intended to be
 Vimspector is not:
 
 * a debugger! It's just the UI and some glue.
-* fast. It's abstractions all the way down. If you want a fast, native debugger, there are other options.
-* comprehensive. It's limited by DAP, and limited by my time. I implement the features I think most users will need, not every feature possible.
-* for everyone. Vimspector intentionally provides a "one size fits all" UI and aproach. This means that it can only provide essential/basic debugging features for a given language. This makes it convenient for everyday usage, but not ideal for power users or those with very precise or specific requirements. See [motivation](#motivation) for more info. 
+* fast. It's abstractions all the way down. If you want a fast, native debugger,
+  there are other options.
+* comprehensive. It's limited by DAP, and limited by my time. I implement the
+  features I think most users will need, not every feature possible.
+* for everyone. Vimspector intentionally provides a "one size fits all" UI and
+  aproach. This means that it can only provide essential/basic debugging
+  features for a given language. This makes it convenient for everyday usage,
+  but not ideal for power users or those with very precise or specific
+  requirements. See [motivation](#motivation) for more info. 
 
 ## Status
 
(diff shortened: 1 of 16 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/756)

puremourning (commented):
> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/756)* status: all files reviewed, 5 unresolved discussions

___
*[`autoload/vimspector.vim` line 90 at r1](https://reviewable.io/reviews/puremourning/vimspector/756#-NTTIr92BVFlQTqJLLUP:-NTTIr932ouBiuQnXt8M:brbzsr5) ([raw file](https://github.com/puremourning/vimspector/blob/3b3b4395b432bce2ea21f53f8082fcb06c1fddab/autoload/vimspector.vim#L90)):*
> ```vim
> s = _vimspector_session_man.FindSessionByName( vim.eval( 'a:name' ) )
> if s is not None:
>   _vimspector_session = _vimspector_session_man.DestroyRootSession(
> ```

use _VimspectorSwitcchTo
___
*[`autoload/vimspector/internal/state.vim` line 52 at r1](https://reviewable.io/reviews/puremourning/vimspector/756#-NTTJARR6ji7EJfl2Z8d:-NTTJARR6ji7EJfl2Z8e:bagmfr) ([raw file](https://github.com/puremourning/vimspector/blob/3b3b4395b432bce2ea21f53f8082fcb06c1fddab/autoload/vimspector/internal/state.vim#L52)):*
> ```vim
>   global _vimspector_session
> 
>   if _vimspector_session is not None:
> ```

do nothing if new session is same as previous
___
*[`autoload/vimspector/internal/state.vim` line 75 at r1](https://reviewable.io/reviews/puremourning/vimspector/756#-NTTJEsx3oFrqF8vzc51:-NTTJEsx3oFrqF8vzc52:bepxwz5) ([raw file](https://github.com/puremourning/vimspector/blob/3b3b4395b432bce2ea21f53f8082fcb06c1fddab/autoload/vimspector/internal/state.vim#L75)):*
> ```vim
> function! vimspector#internal#state#NewSession( options ) abort
>   py3 << EOF
> s = _vimspector_session_man.NewSession( **vim.eval( 'a:options' ) )
> ```

don't use `s` here
___
*[`python3/vimspector/debug_session.py` line 534 at r1](https://reviewable.io/reviews/puremourning/vimspector/756#-NTTJpvX6uNj2us-fK6Q:-NTTJpvX6uNj2us-fK6R:bamzboe) ([raw file](https://github.com/puremourning/vimspector/blob/3b3b4395b432bce2ea21f53f8082fcb06c1fddab/python3/vimspector/debug_session.py#L534)):*
> ```Python
>   def SwitchTo( self ):
>     if self.HasUI():
>       vim.current.tabpage = self._uiTab
> ```

hmmm. I'm not sure we always want to do this when `_VimspectorSwitchTo` is used.

We need to have `_VimspectorMakeActive` and `_VimspectorSwitchTo` with switch-to meaning make active and display on screen and make active meaning just make it the active _vimspector_session? 
___
*[`python3/vimspector/utils.py` line 1091 at r1](https://reviewable.io/reviews/puremourning/vimspector/756#-NTTKZ5L0CUa5BsdzBKC:-NTTKZ5L0CUa5BsdzBKD:b-u9p8mw) ([raw file](https://github.com/puremourning/vimspector/blob/3b3b4395b432bce2ea21f53f8082fcb06c1fddab/python3/vimspector/utils.py#L1091)):*
> ```Python
> def BufferNameForSession( name, session_id ):
>   return f'{name}[{session_id}]'
>   # return name
> ```

remove comment

puremourning (commented):
> Reviewed 5 of 5 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/756)* status: all files reviewed, 2 unresolved discussions

*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/756#-NTTNKov2qB7hF0kJOzK:-NTTNKov2qB7hF0kJOzL:b-jdswez) (no related file):*
just needs tests now

___

puremourning (commented):
:lgtm:

> Reviewed 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/756#-:-NTYRtj123DNhKPt-PjU:bnfp4nl)* status: all files reviewed, 1 unresolved discussion

puremourning (commented):
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/756)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @puremourning)

*[a discussion](https://reviewable.io/reviews/puremourning/vimspector/756#-NTTNKov2qB7hF0kJOzK:-NTYWOEBDb2zEGP2Zxj9:bpj612f) (no related file):*

just needs tests now
</blockquote></details>

will add these later

___

gagara:
Tested with java remote debug. Works perfectly for me. The only one issue I've faced:
Breakpoint labels are lost when switching between sessions. Breakpoints itself are ok, I can see them in the list and debugger stops correctly, but no labels in source code.
Thanks a lot for really cool feature.

puremourning:
> Breakpoint labels are lost when switching between sessions. Breakpoints itself are ok, I can see them in the list and debugger stops correctly, but no labels in source code.
Thanks a lot for really cool feature.

This is intentional currently. We clear the signs for non-current sessions. Initially I didn't do this but I found that without this it was confusing to determine if a breakpoint existed or not. 

Say you had the same source file in multiple root sessions. Having a bp sign there could trick you into thinking the current session would stop there. But if the bonding was for another session. It wouldn't stop. 

It's generally not obvious if breakpoints should be per-vim-instance or per-root-debug-session. Currently they are per root debug session.

puremourning:
Thanks for the feedback @gagara its really valuable.

gagara:
Hmm, in my case they was cleared for all sessions. I was able to see them in breakpoint list only. Is it possible in this case to show that breakpoint is unreachable (yellow) when I switch to another session?

PR #768, opened 2023-05-01
Allow editing of breakpoint options in breakpoints window

### LINKED ISSUE #299: [Feature Request]: Allowing viewing / editing conditional breakpoints
Conditional breakpoints work, but it doesn't appear to be possible to see the conditions already assigned to a breakpoint. Also, when you add a conditional breakpoint where there is already a (conditional or not) breakpoint, it deletes the breakpoint and discards your input. That is rather unexpected behaviour.

Suggest that if you try to add a conditional breakpoint where there already is a breakpoint, the input is pre-populated with the current condition (if any), and of course it first deletes the breakpoint but then also adds the new one; that way you can edit/view the conditional breakpoint by adding another conditional breakpoint, and just press enter again if you just wanted to view it, solving both the deletion and visibility problems without adding new interface elements.

For the viewing of the conditions of breakpoints, an even better solution would be to have a separate window that lists all the locations of (and conditions of) the breakpoints. This would also facilitate navigating to breakpoints, which would also be really useful.

### LINKED ISSUE #737: [Feature Request]: Editing conditional breakpoints' conditions in the breakpoint window
It's probably a niche usecase, but I report it just in case. Recently, in debugging some bugged code I really had little clue about, I had to do the following:
  1. set 1 breakpoint at a given line where a pointer value was available
  2. run the program
  3. when the program stops a that breakpoint, inspect that pointer (say the value is `0xFOO123`)
  4. set _a lot_ of conditional breakpoints around with the condition that `someLocalPtr == 0xFOO123`.

Clearly, from run to run, `0xFOO` would change, so I had to manually reset the condition of every single one of the several breakpoints I set at step 4.

Well, I guess that if I could edit the conditions of breakpoints in the breakpoint window, I could use all the power of Vim, such as visual block editing, `.`, macros, ...

It's not a scenario that happens frequently, so the alternative I have is to become quick at <kbd>Ctrl+P</kbd> to set for a conditional breakpoint the same condition as the previous one.

**Additional context**

puremourning:
…inbar in neovim

Fixes #299 
Fixes #737

### DIFF
--- a/README.md
+++ b/README.md
@@ -910,12 +910,14 @@ The following mappings apply by default in the breakpoints window:
 * `t`, `<F9>` - toggle, i.e. enable/disable breakpoint
 * `T` - toggle, i.e. enable/disable ALL breakpoints
 * `dd`, `<Del>` - delete the current breakpoint
+* `cc`, `C` - edit the current breakpoint options
 * `i`, `a`, `o` - add a new line breakpoint
 * `I`, `A`, `O` - add a new function breakpoint
 * `<Enter>` or double-click - jump to the line breakpoint
 
 A WinBar is provided (where supported) too. This adds functions like
-saving/restoring sessions and clearing all breakpoints too.
+saving/restoring sessions, clearing all breakpoints, and resetting the exception
+breakpoints options.
 
 ### Line breakpoints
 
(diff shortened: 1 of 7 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/768)

puremourning (commented):
:lgtm:

> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/768#-:-NUMw5hM9tfaTa2hLz3R:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @puremourning)

PR #777, opened 2023-06-14
Expand references in calculus function arguments

puremourning:
This allows for the use of variables as arguments to functions.

In particular, the following is now possible:

     "configuration": {
       "request": "attach",
       "processId": "${PickProcess(\"${programName}\")}",
       "program": "$BUILD/platform/bin64/${programName}",
       "expresssions": "native"
     }

The result is that you're asked for the program _name_ and it filters for that, and uses it as the binary.

Doesn't work for function-calls-inside-function-calls because of not enough escaping levels for parameters, but that's probably OK. The above is a real use case however, and it works.

This is mad, gnarly stuff which probably shouldn't be used, but that's also true of the whole of this entire macro-templating system, which honestly needs to be compeltely ditched!

### DIFF
--- a/python3/vimspector/utils.py
+++ b/python3/vimspector/utils.py
@@ -672,7 +672,15 @@ def ExpandReferencesInString( orig_s,
       key = e.name + e.arg_hash
 
       if e.name in calculus:
-        mapping[ key ] = calculus[ e.name ]( *e.args )
+        # Expand any recursive mappings in the args.... eeeek
+        new_args = []
+        for arg in e.args:
+          new_args.append( ExpandReferencesInObject( arg,
+                                                     mapping,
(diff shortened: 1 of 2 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/777)

puremourning (commented):
:lgtm:

> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/puremourning/vimspector/777#-:-NXvTqYy4wsEZprRIFuD:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @puremourning)

PR #791, opened 2023-08-18
Fix some typos in bug issue template

Aster89:
- reproducable -> reproducible
  - "Works in vim?" seems a weird question, considering one is doing a bug report, e.g. it looks like you're asking "does the example to show that something doesn't work, work?"


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/791)

puremourning:
Thanks!

PR #827, opened 2024-01-17
[python/debugpy] update sha256 for debugpy

xingchensong:
image

without `checksum`, we have to redownload debugpy.zip


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/827)

puremourning:
The reason the checksum was removed is that it is not stable and kept causing issues. So unless it is now stable I don’t want to revert to that situation.

xingchensong:
got it

PR #857, opened 2024-05-06
Stop building release tarballs

puremourning:
From issues, it seems nobody really uses them, and they have a cost in terms of CI time.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/857)

PR #876, opened 2024-08-18
Connection to gadgets on IPv6-enabled host is failing

lelutin:
On hosts where IPv6 is enabled and preferred by the system, gadgets might be listening to ::1 instead of 127.0.0.1. In those situations, vimspector times out, unable to connect to the gadget even though they've both agreed about which port to use.

Changing the default host value to 'localhost' makes it possible for IPv6-enabled hosts to use the same default in vimspector than in the gadgets.

### DIFF
--- a/autoload/vimspector/internal/channel.vim
+++ b/autoload/vimspector/internal/channel.vim
@@ -88,7 +88,7 @@ function! vimspector#internal#channel#StartDebugSession(
           \              )
   endif
 
-  let l:addr = get( a:config, 'host', '127.0.0.1' ) . ':' . a:config[ 'port' ]
+  let l:addr = get( a:config, 'host', 'localhost' ) . ':' . a:config[ 'port' ]
 
   echo 'Connecting to ' . l:addr . '... (waiting for up to 10 seconds)'
   " FIXME: This _always_ waits 10s; the neochannel version is quicker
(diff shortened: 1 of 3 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/876)

puremourning:
Thanks for the PR, but unfortunately, we force the use of IPv4 for the exact opposite reason! See 4206d0e5 for example.

There were a swathe of issues when using `localhost` because for some unspeakable reason certain OS distributions _default_ to using IPv6 and so just hang when using `localhost`, and _most_ servers are using IPv4 for this comms.

I think if your server uses IPv6 by default, you just need to set host to `::1` when connecting.

lelutin:
Oh I see, thanks for the details. It's a bit unfortunate to still be defaulting to ipv4 nowadays but I can understand that ipv6 has still not established its hegemony and so ipv4 keeps on being the safe default.

So instead of this MR, maybe we could add a quick mention of the default value (and thus address family) for `host` somewhere?
The most relevant place I could find was in https://puremourning.github.io/vimspector/configuration.html#debug-adapter-configuration but we could also drop a note in https://github.com/puremourning/vimspector/wiki/Additional-Language-Support#gadget-installer-files

PR #918, opened 2025-05-24
seq =1 in first message

MrBoboGet:
The DAP protocol (https://microsoft.github.io/debug-adapter-protocol/specification) requires that the first message has sequence number equal to 1. This requirement does not seem to be widely enforced however, but it does create a problem when using the latest version of lldb-dap directly, as it simply aborts the connection, making it completely unusable.


puremourning:
@mergifyio rebase

puremourning:
Oh hey, thanks for this. Sorry somehow I completely missed it.

It LGTM

PR #932, opened 2025-08-04
Update README.md with Lazy.nvim plugin manager info

Djedouas:
I had quite some trouble to finally load properly vimspector with [Lazy.nvim](https://github.com/folke/lazy.nvim) package manager.

Running `:Lazy load vimspector` once, after Vimspector having been installed by Lazy.nvim, make all Vimspector functions and features available.

I guess that it sources vimspector's `autoload/` directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,8 +227,16 @@ There are 2 installation methods:
    Plugin 'puremourning/vimspector'
    ```
 
-2. Install some 'gadgets' (debug adapters) - see [here for installation commands](#install-some-gadgets) and [select gadgets to install](#supported-languages)
-3. Configure your project's debug profiles (create `.vimspector.json`, or set
+   For Lazy.nvim, use:
+
+   ```vim
+   return { "puremourning/vimspector" }
+   ```
+
+   and launch `:Lazy load vimspector` for the first start.
+
+3. Install some 'gadgets' (debug adapters) - see [here for installation commands](#install-some-gadgets) and [select gadgets to install](#supported-languages)
+4. Configure your project's debug profiles (create `.vimspector.json`, or set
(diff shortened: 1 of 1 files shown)


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/932)

puremourning:
Thanks for sharing, but documenting the idiosyncratic broken behaviour of broken plugin systems is not really for each and every plugin to do…

Either vimspector is broken or the plugin system is. Do we know if there is something broken in vimspector here? I’d rather stuff worked than document workarounds.

Djedouas:
I think that vimspector is not broken.

This is probably due to the way Lazy.nvim plugin manager works.

Extract from [the documentation](https://lazy.folke.io/usage#%EF%B8%8F-startup-sequence):

> lazy.nvim does NOT use Neovim packages and even disables plugin loading completely (vim.go.loadplugins = false). It takes over the complete startup sequence for more flexibility and better performance.

and reading further on the same documentation page, I found:

<table><thead><tr><th>Command</th><th>Lua</th><th>Description</th></tr></thead><tbody><tr><td><code>:Lazy load {plugins}</code></td><td><code>require("lazy").load(opts)</code></td><td>Load a plugin that has not been loaded yet. Similar to <code>:packadd</code>. Like <code>:Lazy load foo.nvim</code>. Use <code>:Lazy! load</code> to skip <code>cond</code> checks.</td></tr></tbody></table>

which evokes packer, that's why I tried this command, which worked.

puremourning:
Ok well the docs for the plugin system cover this case so I don’t need to repeat it here. Otherwise I’d have an entry for every quirky plugin system :)

thanks for taking the time to send a PR though, it’s mich appreciated!

PR #934, opened 2025-08-09
Allow editing condition on function and data breakpoints

samisalreadytaken:
No new tests because it's a basic copy-paste job, and it works while connected and disconnected to a server.


puremourning:
Do We also need to update the docs?

samisalreadytaken:
You have the comment "Can't find an adapter that supports conditional function breakpoints" in `breakpoints.test.vim` from 5 years ago. I'll check if they do now when I have time, but what should the approach be if existing test servers still don't support them?

> Do We also need to update the docs?

I don't see any mention of breakpoint editing in the docs. But ["Supported debugging features - conditional breakpoints (function, line)"](https://github.com/puremourning/vimspector/blob/a1ee02d4e5df05d411e93740465db1177ba42b4c/doc/vimspector.txt#L281) line could include "data" as well.

Irrelevant, though, [the description of `hitCondition`](https://github.com/puremourning/vimspector/blob/a1ee02d4e5df05d411e93740465db1177ba42b4c/doc/vimspector.txt#L1288) is misleading as it is perfectly valid to have `condition` and `hitCondition` together.

samisalreadytaken:
I use conditional function and data breakpoints with my debugger. If there is no test for function breakpoint conditions (which could already be added but not edited) already, I'm confused as to what exactly I'm supposed to add tests for. Just the window output but not the functionality?

This patch simply adds code path for `type == 'F'` and `type == 'D'` for the `Edit` button (`cc`) in breakpoints view.

PR #939, opened 2025-08-26
Log on Win32 with WSL

samisalreadytaken:
If the Windows system has WSL installed, which is not unlikely on a developer's machine, use `tail` through it.


puremourning:
This change is [Reviewable](https://reviewable.io/reviews/puremourning/vimspector/939)

